PR #22, opened 2014-11-14
add executable to handle leave actions

dbluelle:
new attempt for https://github.com/lxde/lxqt/issues/319

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,3 +336,6 @@ set(CPACK_GENERATOR TBZ2)
 set(CPACK_SOURCE_GENERATOR TBZ2)
 set(CPACK_SOURCE_IGNORE_FILES /build/;.gitignore;.*~;.git;.kdev4;temp)
 include(CPack)
+
+# lxqt-leave executable
+add_subdirectory(lxqtleave)
--- a/lxqtleave/CMakeLists.txt
+++ b/lxqtleave/CMakeLists.txt
@@ -1,0 +1,35 @@
+cmake_minimum_required(VERSION 2.8.11)
+project(lxqt-leave)
+
+set(CPP_FILES
+    main.cpp
+)
+
+
+# Translations **********************************
+include(LXQtTranslate)
+lxqt_translate_desktop(DESKTOP_FILES
+    SOURCES
+    resources/lxqt-hibernate.desktop.in
+    resources/lxqt-lockscreen.desktop.in
+    resources/lxqt-logout.desktop.in
+    resources/lxqt-reboot.desktop.in
+    resources/lxqt-shutdown.desktop.in
+    resources/lxqt-suspend.desktop.in
+)
+#************************************************
+
+
+add_executable(lxqt-leave
+    ${CPP_FILES}
+    ${DESKTOP_FILES}
+)
+
+target_link_libraries(lxqt-leave
+    ${LXQT_LIBRARIES}
+    ${LXQT_LIBRARY_NAME}
+)
+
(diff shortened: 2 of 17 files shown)


antis81:
If you ask me, yes :smile:. But I'm just a user.

jleclanche:
Due to lack of feedback from others, I'm pushing for this. There's been a lot of back and forth on this and we need some form of it in 0.9 anyway.

Landed in 63cf48a.

PCMan:
Maybe lxqt-session is a better place to put this command?
It's reasonable for a session manager to provide a tool for quitting the
session.
We used to do this in lxde.
2015/1/12 下午3:33 於 "Jerome Leclanche" <notifications@github.com> 寫道：

jleclanche:
I want this in lxqt-session, but someone (I think it was you, actually) argued against it because they didn't want users to be locked into lxqt-session.

PCMan:
2015/1/12 下午3:55 於 "Jerome Leclanche" <notifications@github.com> 寫道：
>
> I want this in lxqt-session, but someone (I think it was you, actually)
argued against it because they didn't want users to be locked into
lxqt-session.
It was me? Ouch!
Well, this can be useful as a standalone program.
However, if nobody wants to maintain a separate package, we need some
tradeoff.
Then lxqt-session is probably the best place for the tool.

jleclanche:
I'm +1 on moving it to lxqt-session. Let's wait on other people's feedback but this needs to happen before the end of the week if it happens. ping @surlykke @paulolieuthier @luis-pereira

Vladimir-csp:
I vote for modularity, but i`m just a user here.

jleclanche:
This wouldn't be a problem if we could eg. rely on systemd being present, seeing as systemd has APIs which handle this so we could just rely on this. My suggestion if you do use a different session manager and for whatever reason you *need* those shortcuts, you should use a systemd call in the desktop file. If you don't have systemd, there are other wrappers you can install...

paulolieuthier:
I'm actually happy with both solutions. @PCMan's arguments before and now are strong.

agaida:
i think that the session is the best place for it - i don't like micro packages.

jleclanche:
@PCMan we have enough agreement to justify the move imo. Can you take care of it? Or @agaida?

PCMan:
Assign to me if no other people can do it.
I don't have time for coding now but can handle it next week.
2015/1/12 下午7:17 於 "Jerome Leclanche" <notifications@github.com> 寫道：

jleclanche:
@agaida said on IRC he can pick it up.

agaida:
fine, will do that

dbluelle:
I think the main reason for _not_ putting this into lxqt-session was that if we do that,
lxqt-panel may need a hard dependency on lxqt-session.
(see https://github.com/lxde/lxqt-panel/pull/93 and https://github.com/lxde/lxqt/issues/319 )
lxqt-panel will not have any "leave" buttons without the desktop files,
or am I missing something here?

agaida:
@dbluell - no, not really - if the session isn't installed, there is no desktop-file nor executable - so what? Maybe i miss something too :)

jleclanche:
It's a soft dependency, not a hard dependency.

dbluelle:
>  The only thing is - you cant place these things into a quick-starter, the start menu functionality should not be touched.

Does that mean we should drop https://github.com/lxde/lxqt-panel/pull/93 ?
Without that PR it will still not be possible to drag/drop the leave actions into a quick-starter 
(which is the whole point why I started this... )

PR #31, opened 2015-01-15
remove lxqtleave from liblxqt

agaida:
and put it into lxqt-session. refs #22


antis81:
All really good stuff :+1:. Just ... how can I call the shutdown dialog?

As a user with a "customized" environment (cairo-dock instead of lxqt-panel), I currently have two options:

1. Open a terminal and write something like `lxqt-leave --logout`.
2. Open lxqt-runner by pressing `Alt+F2` and chose from the menu.

Did I miss something?

jleclanche:
lxqt-leave is just a wrapper for various session logout mechanisms (eg systemd). you can just use the one your distro gives you, you dont have to use lxqt-leave itself.

antis81:
Thus, I can re-install lxsession-logout for Lubuntu? Yeah, I try that ... thanks!

PR #33, opened 2015-01-19
Workaround Qt 5.4 issue: Re-add a config file to the file watcher if it's already deleted to get proper change notifications.

PCMan:
This is to fix lxde/lxqt#441 - [Regression] All LXQt::Settings and QSettings file change monitoring stop working.
https://github.com/lxde/lxqt/issues/441

### DIFF
--- a/lxqtsettings.cpp
+++ b/lxqtsettings.cpp
@@ -34,6 +34,7 @@
 #include <QMutex>
 #include <QFileSystemWatcher>
 #include <QSharedData>
+#include <QTimerEvent>
 
 #include <XdgIcon>
 #include <XdgDirs>
@@ -44,6 +45,7 @@ class LxQt::SettingsPrivate
 {
 public:
     SettingsPrivate(Settings* parent):
+        mChangeTimer(0),
         mParent(parent)
     {
     }
@@ -51,6 +53,7 @@ public:
     QString localizedKey(const QString& key) const;
 
     QFileSystemWatcher mWatcher;
+    int mChangeTimer;
 
 private:
     Settings* mParent;
@@ -106,7 +109,7 @@ Settings::Settings(const QString& module, QObject* parent) :
         sync();
     }
     d_ptr->mWatcher.addPath(this->fileName());
-    connect(&(d_ptr->mWatcher), SIGNAL(fileChanged(QString)), this, SLOT(fileChanged()));
+    connect(&(d_ptr->mWatcher), &QFileSystemWatcher::fileChanged, this, &Settings::_fileChanged);
 }
 
 
@@ -125,7 +128,7 @@ Settings::Settings(const QString &fileName, QSettings::Format format, QObject *p
         sync();
     }
     d_ptr->mWatcher.addPath(this->fileName());
-    connect(&(d_ptr->mWatcher), SIGNAL(fileChanged(QString)), this, SLOT(fileChanged()));
+    connect(&(d_ptr->mWatcher), &QFileSystemWatcher::fileChanged, this, &Settings::_fileChanged);
 }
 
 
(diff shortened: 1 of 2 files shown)


PCMan:
@luis-pereira , @paulolieuthier , @jleclanche Please help review.
Thanks!

jleclanche:
Looks good at a glance, didn't test.

Did you file a bug/ask about it upstream?

PCMan:
@jleclanche No, because the QSettings change is a feature, not a bug.
The QSettings API have never guaranteed that it won't write to a temp file.
They did not break any public API, but we rely on its implementation detail.
Hence LXQt::Settings is broken. To be fair, however, it's our problem.

PCMan:
In my PR, I also add a 100ms delay before emitting the settingsChange() signal.
This can compress rapid serial changes to the same file into one notification and avoid emitting the signal too many times unnecessarily. I'm not sure if this "feature" will break existing code so I ask for you guys to review the PR. Thanks!
If it's good to merge, this can also fix lxde/lxqt#441.

PCMan:
Another important thing for the PR is to ensure that it won't break Qt 5.3.
The fix works for me, but your tests will be appreciated.

agaida:
@PCMan - built liblxqt with your fix - seems to be fine with Qt 5.3.2 in a installed system

luis-pereira:
@PCMan I'm using Qt 5.4 on Arch. The Widget Style and Font aren't working.
Interestingly LXQtPlatformTheme::onSettingsChanged() is only called one time (when lxqt-config-appearance starts).

paulolieuthier:
Running Qt 5.4 on Arch as well. Everything works nicely now, except for the "LXQt Theme" that doesn't change.

PCMan:
Both lxqt-qtplugin and liblxqt should be fixed at the same time.
@luis-pereira are you using the latest "lxqt-qtplugin"? It should be fixed in the latest lxqt-qtplugin.
The "onSettingsChanged() is only called once" problem is exactly caused by the bug, but I aready added a workaround for it and it works for me. Can you try it again?
@paulolieuthier LXQt theme is handled by liblxqt, not lxqt-qtplugin.
So if you're using the latest lxqt-qtplugin, everything other than lxqt theme should work.
You have to apply my fix mentioned in the liblxqt PR to workaround the problem.
Are you trying the PR?

PCMan:
Other existing option for the settings are:
1. Use the feature-rich GFileMonitor from glib to solve the problem and do not use the "broken" QFileSystemWatcher API. (Since glib is installed on every system, this option does not add new dependencies)
2. Use KConfig library from KF5 to replace LXQt::Settings (ini file based, provides notifications, but complicated and has unknown performance)
3. Use gsettings from glib as the backend of LXQt::Settings (can reuse glib, but this requires dconf by default. There is a ini file based backend, though)

luis-pereira:
@PCMan I wasn't running the patched lxqt-qtplugin. It's working properly now. :smiley:
Option 1 seems to be better option IMO. But, I didn't "exhaustively" analyzed it.

jleclanche:
For 0.9, any large change is obviously out of the question.

I'm -1 on GFileMonitor. Unnecessarily complex, we might as well use a proper config api.

On paper it would make sense to use KConfig but then KConfig has to actually be good. And it can't pull in too many unrelated frameworks, either... the KF5 dependency tree is still a mess and KDE devs don't seem to have an incentive to work on that. I don't know how it is for KConfig.

From various chats I had with @desrt a while back, it would seem that gsettings is an excellent option. Is anyone up for looking into pros and cons between kconfig and gsettings?

PCMan:
@jleclanche We may do more study after 0.9.
Maybe open another issue for it and close the current one if the workaround works?

jleclanche:
Ok.

PR #34, opened 2015-01-19
Portuguese update

smarquespt:


luis-pereira:
@smarquespt Thanks. Landed on e73e723a6cd1444ec4106dc8671ffd1

PR #36, opened 2015-02-02
- Remove KF5 from prefix. This will fix compilation when KDE framework h...

heliocastro:
...eaders are not in standard paths.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ find_package(Qt5Widgets REQUIRED QUIET)
 find_package(Qt5DBus REQUIRED QUIET)
 find_package(Qt5X11Extras REQUIRED QUIET)
 find_package(Qt5LinguistTools REQUIRED QUIET)
-find_package(Qt5Xdg REQUIRED)
+find_package(Qt5Xdg REQUIRED QUIET)
 find_package(KF5WindowSystem REQUIRED QUIET)
 message(STATUS "Building with Qt ${Qt5Core_VERSION_STRING}")
 
--- a/lxqtsingleapplication.cpp
+++ b/lxqtsingleapplication.cpp
@@ -27,8 +27,8 @@
 
 #include "lxqtsingleapplication.h"
 #include "singleapplicationadaptor.h"
-#include <KF5/KWindowSystem/KWindowSystem>
-#include <KF5/KWindowSystem/NETWM>
+#include <KWindowSystem/KWindowSystem>
+#include <KWindowSystem/NETWM>
 #include <QDBusMessage>
 #include <QWidget>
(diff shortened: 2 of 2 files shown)


paulolieuthier:
But it's not optional, it's used in several components of lxqt.

agaida:
as i understand this: find_package(KF5WindowSystem REQUIRED QUIET
the package is found, no KF5/ is needed. I may be wrong.
lgtm

heliocastro:
Is, the package is found, but this doesn't mean that the include headers path points to proper place.
For example, buy self build KDE is located under /opt/kf5, so header include path points to /opt/kf5/KF5 , so it will never works.
The fact that worked on standard path is that /usr/include is always a include PATH, so, even the cmake KWindowSystem setting path to /usr/include/KF5, the /usr/include is there, those working on previous code.

luis-pereira:
@paulolieuthier You are right, ```#include <KF5/KWindowSystem/...>``` is used in many places. Probably every single place that uses KF5, uses ```#include <KF5/KWindowSystem/...>```
By looking at this code from KF5WindowSystemTargets.cmake configured with: -CMAKE_INSTALL_PREFIX=/usr
```
set_target_properties(KF5::WindowSystem PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/KF5/KWindowSystem;/usr/include;/usr/include;/usr/include/KF5;${_IMPORT_PREFIX}/include/KF5"
  INTERFACE_LINK_LIBRARIES "Qt5::Widgets"
)
```
and this one configured with -DCMAKE_INSTALL_PREFIX=/opt
```
set_target_properties(KF5::WindowSystem PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/KF5/KWindowSystem;/opt/include/KF5;${_IMPORT_PREFIX}/include/KF5"
  INTERFACE_LINK_LIBRARIES "Qt5::Widgets"
)
```
I do agree with @heliocastro: ```#include <KWindowSystem/...........>``` always work and ```#include <KF5/KWindowSystem/.......>``` fails for non-standard locations.

@heliocastro GTM. But as @paulolieuthier said that's a lot more to fix :smirk:

heliocastro:
Yes, I just have the pull request per feature, and will be compiling other modules as well today to find the issues.

paulolieuthier:
Oh, I misunderstood it. If it works for everyone, let's do it.

luis-pereira:
@heliocastro Thanks. This kind of stuff makes the release process smoother.

agaida:
@luis-pereira @heliocastro @paulolieuthier - could you please review #18 #19 #20 above?

PR #38, opened 2015-03-24
addplugindialog: support for managing plugins

palinek:
references lxde/lxqt#536, lxde/lxqt#552 (and maybe others)
coupled with (must be merged first) lxde/lxqt-panel#160

Added viewing of delegated context menu for each plugin type:
- in list user can invoke context menu (right-click, 'menu' key) to manage all plugins of current type

Issues/questions:
- the 'Add widget' button now may be obsolete
- the context menu may be a bit tricky for users -> should we add a note somewhere about it?
- i had just a quick search if the dialog is used elsewhere in lxqt and found nothing but `lxqt-panel`. Should we be backward compatible (=> leave old overloads for `pluginAdded`, `pluginRemoved` & `setPluginsInUse`)

### DIFF
--- a/addplugindialog/lxqtaddplugindialog.cpp
+++ b/addplugindialog/lxqtaddplugindialog.cpp
@@ -33,6 +33,7 @@
 #include <XdgIcon>
 #include <QListWidgetItem>
 #include <QIcon>
+#include <QMenu>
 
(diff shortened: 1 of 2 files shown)


palinek:
...merging to get more feedback

PR #39, opened 2015-04-16
gridlayout: avoid empty margins when stretching is active

palinek:
In case stretching is active and for cells rows/cols * itemHeight/itemWidth != height/width
not all possible space was used. This can cause that we are not able to properly style
elements see lxde/lxqt#597

closes lxde/lxqt#597

### DIFF
--- a/lxqtgridlayout.cpp
+++ b/lxqtgridlayout.cpp
@@ -542,7 +542,7 @@ void GridLayout::setGeometry(const QRect &geometry)
             itemWidth = d->mCellSizeHint.width();
         }
 
-        itemWidth = qBound(d->mPrefCellMinSize.width(), itemWidth, d->mPrefCellMaxSize.width());
+        itemWidth = qBound(qMin(d->mPrefCellMinSize.width(), maxX), itemWidth, d->mPrefCellMaxSize.width());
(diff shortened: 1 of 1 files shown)


palinek:
@paulolieuthier, @luis-pereira could you have a look on this?

The result is that not all cells have the same size in case count * size != totalSize => but the difference is max 1px

paulolieuthier:
What should we see different?

palinek:
Example:
- panel on bottom
- panel: width 98px, rows 3, icon size 50px
- check the height of each taskbar button by dom plugin
(also make sure, you are using the latest lxqt-common themes -> with no extra margins)

With master you get rows with heights 32px, 32px, 32px => 2px is spare is on the bottom.
With layout-stretch you get rows with heights 33px, 33px, 32px => no spare on the bottom

paulolieuthier:
Works really well when panel is on the bottom. When on the left, a lot of space is unused.

palinek:
>  When on the left, a lot of space is unused.

This is a desired behaviour, not? Vertical stretching is not enabled when panel is vertical.

paulolieuthier:
I mean horizontal stretch when panel is vertical.

PR #42, opened 2015-05-15
Updated german translation

ThomasVie:


pmattern:
The actual commits proposed by this PR have already been merged, the number of files changed is zero.
Don't know what happened or what is intended by like this the PR doesn't seem to make sense.

ThomasVie:
You are right, don't know why I re-opened it.

PR #45, opened 2015-05-26
Sort translations in .desktop files

ThomasVie:
For me it is annoying to search for a locale if the .desktop files are not sorted. This patch will do it.
Additionally fixed some typos.

### DIFF
--- a/translations/liblxqt.ts
+++ b/translations/liblxqt.ts
@@ -132,49 +132,35 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../lxqtscreensaver.cpp" line="70"/>
-        <location filename="../lxqtscreensaver.cpp" line="78"/>
-        <location filename="../lxqtscreensaver.cpp" line="86"/>
-        <location filename="../lxqtscreensaver.cpp" line="98"/>
-        <location filename="../lxqtscreensaver.cpp" line="106"/>
-        <location filename="../lxqtscreensaver.cpp" line="114"/>
-        <location filename="../lxqtscreensaver.cpp" line="123"/>
-        <source>Screen Saver Activation Error</source>
+        <location filename="../lxqtscreensaver.cpp" line="76"/>
+        <source>Screen Saver Error</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../lxqtscreensaver.cpp" line="71"/>
-        <source>An error occurred starting screensaver. xdg-screensaver cannot be started due its crash.</source>
-        <translation type="unfinished"></translation>
-    </message>
-    <message>
-        <location filename="../lxqtscreensaver.cpp" line="79"/>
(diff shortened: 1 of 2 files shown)


cmake/modules/LXQtTranslateDesktop.cmake
```diff
@@ -72,7 +72,7 @@ function(lxqt_translate_desktop _RESULT)
         if (_translations)
             add_custom_command(OUTPUT ${_outFile}
                 COMMAND grep -v "'#TRANSLATIONS_DIR='" ${_inFile} > ${_outFile}
-                COMMAND grep -h ${_pattern} ${_translations} >> ${_outFile}
+                COMMAND grep -h ${_pattern} ${_translations} | sort >> ${_outFile}
```
palinek: It's a matter of taste, but IMO it's better to have all keys for one language together...

ThomasVie:
So, well, do you know a way to sort the file names ${_translations}?

palinek:
Well, we have 2 options (as I see it):
- sort the list of files `_translations`:
```diff
--- a/cmake/modules/LXQtTranslateDesktop.cmake
+++ b/cmake/modules/LXQtTranslateDesktop.cmake
@@ -67,6 +67,7 @@ function(lxqt_translate_desktop _RESULT)
             ${_translationDir}/${_fileName}_*${_fileExt}
             ${_translationDir}/local/${_fileName}_*${_fileExt}
         )
+        list(SORT _translations)
 
         set(_pattern "'\\[.*]\\s*='")
         if (_translations)
```
, but there is a problem that files are searched in 2 directories. @luis-pereira why is also the `local` subdirectory searched (is it used anywhere?)?

- or instruct `sort` to sort based on key in [ ]:
```diff
--- a/cmake/modules/LXQtTranslateDesktop.cmake
+++ b/cmake/modules/LXQtTranslateDesktop.cmake
@@ -72,7 +72,7 @@ function(lxqt_translate_desktop _RESULT)
         if (_translations)
             add_custom_command(OUTPUT ${_outFile}
                 COMMAND grep -v "'#TRANSLATIONS_DIR='" ${_inFile} > ${_outFile}
-                COMMAND grep -h ${_pattern} ${_translations} >> ${_outFile}
+                COMMAND grep -h ${_pattern} ${_translations} | sort -k2 -t[ >> ${_outFile}
                 COMMENT "Generating ${_fileName}${_fileExt}"
             )
         else()
```
, but this has the problem that the key for sorting are characters from [ to the end of line (so for each translation the keys Name, Comment ... are not in the same order)

luis-pereira:
@ThomasVie The place to localize is lxqt-component_locale.desktop. The lxqt-component.desktop is for machine consumption.
@palinek It came from razor. It's an artifact from that time. Soon to be removed. I'm on updating the build system.

ThomasVie:
> The lxqt-component.desktop is for machine consumption.
Yes, I know. But it would be nice to have a reproducible build - so that you and I
produce the same product and all files were bytewise identical.
To sort would be one possibility to achieve it.

PR #46, opened 2015-05-27
Updated .ts files

ThomasVie:
This updates the template and the german translation (removes vanished strings, updated line numbers).


jleclanche:
This isn't right. It's also not necessary. Template updates happen after string freeze.

PR #49, opened 2015-06-20
Sort .desktop translations.

ThomasVie:
Accidentially closed my previous pull request. Here it is refreshed.
Even if the .desktop files are for machine consumption only I prefer the languages sorted.
This also provides identical build results which don't depend on physical file order.


palinek:
That's what I've proposed in the prevoius one, right?

I'm OK with that, but let's leave the decision to @luis-pereira

luis-pereira:
I'm Ok with it. Pls update the commit message to include:
* Fix typos in the comments
* Remove local dir artifact

after that it's GTM.Merge it.

ThomasVie:
Modified comment as suggested.

luis-pereira:
@ThomasVie Thanks

PR #51, opened 2015-06-22
Targets

luis-pereira:
Update the build infrastructure to use CMake Targets.


paulolieuthier:
I'm not a CMake expert. For what I can tell, GTM.

luis-pereira:
@paulolieuthier And it works in an superbuild mode. I'm on lxqt-globalkeys now.

PR #52, opened 2015-06-24
LxQt::Application: unix signal handling & quit

palinek:
references lxde/lxqt#714

### DIFF
--- a/lxqtapplication.cpp
+++ b/lxqtapplication.cpp
@@ -44,9 +44,13 @@ using namespace LxQt;
 #define QAPP_NAME qApp ? qApp->objectName().toUtf8().constData() : ""
 
 #include <cstdio>
-#include <cstdlib>
 #include <unistd.h>
+#include <cstring>
+#include <csignal>
+#include <sys/socket.h>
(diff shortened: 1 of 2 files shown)


luis-pereira:
@palinek @palinek Thinking out loud here:  How do we allow the user to run it's code ?

Sorry for the delays. My time has been taken by the Targets stuff and my laptop that just surrender it's soul to the creator :cry:

palinek:
> Thinking out loud here: How do we allow the user to run it's code ?

Do we want also some universal mechanism to allow children of `LXQt::Application` class install/execute it's own logic for handling (any) signals?

luis-pereira:
@palinek IMO yes. It's the app developer that knows what should be done. What's your opinion ?

palinek:
@luis-pereira something like in @a1fe5a4 ?

luis-pereira:
@palinek Yes. Something like that :smile: Only took a quick look

palinek:
...squashed... is it mergable now?

luis-pereira:
@palinek IMO Yes.

PR #53, opened 2015-07-05
fix build with qt 5.5

rezso:
(lxqtnotification.h:43:5: error: 'Q_OBJECT' does not name a type)


paulolieuthier:
Merged in 7303ea207de0771d6f450a31ec4a1ce69202869b. Thanks.

PR #55, opened 2015-07-22
Removes Cannot open qss file

luis-pereira:
There's lot's of components that don't have an theme file. There's no need
to issue a warning, not even an debug message.

### DIFF
--- a/lxqtsettings.cpp
+++ b/lxqtsettings.cpp
@@ -482,7 +482,6 @@ QString LxQtThemeData::loadQss(const QString& qssFile) const
     QFile f(qssFile);
     if (! f.open(QIODevice::ReadOnly | QIODevice::Text))
     {
-        qWarning() << "Theme: Cannot open file for reading:" << qssFile;
         return QString();
     }
 


paulolieuthier:
GTM. This is annoying.

pmattern:
Hasn't this already been fixed as part of lxde/lxqt#666? At least I can't see it any longer with lxqt-config-{monitor,notificationd,session} and lxqt-{leave,openssh-askpass} which all were affected in earlier versions.

palinek:
> Hasn't this already been fixed in lxde/lxqt#666 (comment)? At least I can't see it any longer with lxqt-config-{monitor,notificationd,session} and lxqt-{leave,openssh-askpass} which all were affected in earlier versions.

Indeed, it was "removed" for all build types except `Debug` in lxde/liblxqt@b16807af8188ce9d5017fa2dff84b0f5f8d14ce3 & lxde/liblxqt@f8a1ecaf60c9e12f97fe16f8c19c9a4f428a6356. When removing this kind of debugs we've had a talk on IRC with @jleclanche about what message should leave or not...

luis-pereira:
That's an no added value debug message. It's not an warning for sure. And even as an commit message what debug value it provides ? None, IMO.

jleclanche:
ship it, it's useless

PR #57, opened 2015-08-23
Update Greek (el) translation

dglent:
Update Greek translation

### DIFF
--- a/translations/liblxqt_el.ts
+++ b/translations/liblxqt_el.ts
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="el_GR">
+<TS version="2.1" language="el">
 <context>
     <name>AddPluginDialog</name>
     <message>
         <source>Add plugins</source>
-        <translation type="obsolete">Προσθήκη επεκτάσεων</translation>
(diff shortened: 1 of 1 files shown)


jleclanche:
What did you use to generate this? The output looks wrong.

dglent:
I didn't generate it. I 've just update the translations of the current file <liblxqt_el_GR.ts> 
I used Lokalize and the i opened it also in qt5linguist to check if there are any errors and it was ok.

jleclanche:
When you update it in a tool, the xml gets regenerated. This particular tool unescaped `&gt;`, changed quoting and attribute ordering...

Weird, anyway, gtm

pmattern:
What about attribute "vanished"?
To my knowledge it is used to tag elements that are no longer needed due to changes in the actual source code. The idea is to avoid translation work that has already been done getting lost by just tagging instead of deleting from the TS files. So no point in translating those items, AFAIK.

dglent:
I think that Lokalize has removed the tags (i don't know why) but normally it is not a problem. 
I didn't have any problem with translations and i translate all ts files with Lokalize in order to use the spell check, the TM and the Glossary which i have collated since many years.
They should be tagged again in next lupdate run.

dglent:
I updated the commit and i removed the country specific.
Saving with qt5linguist it has fixed some changes in xml syntax as the place of the 'line=' at the end of the line of the 'location' tag

PR #58, opened 2015-08-25
Lock screen on suspend and hibernate

mid-kid:
Fixes lxde/lxqt#775.

The title pretty much says all.
This patch used to be a personal quick-fix, and as such, I don't know if this is the proper way to do it. Neither do I know for sure if this behaviour is acutally expected, or should be configureable or something.
Also, this patch has only been tested on liblxqt-0.9.0, and I haven't tested it on current master, yet I see no reason why it should work any different (secretly I'm too lazy to recompile libqtxdg to master).

### DIFF
--- a/lxqtpower/lxqtpower.cpp
+++ b/lxqtpower/lxqtpower.cpp
@@ -34,8 +34,10 @@
 using namespace LxQt;
 
 Power::Power(QObject *parent) :
-    QObject(parent)
+    QObject(parent),
+    mScreenSaver(this)
 {
+    connect(&mScreenSaver, SIGNAL(done()), &mLoop, SLOT(quit()));
     mProviders.append(new CustomProvider(this));
     mProviders.append(new SystemdProvider(this));
     mProviders.append(new UPowerProvider(this));
@@ -66,11 +68,18 @@ bool Power::doAction(Power::Action action)
 {
     foreach(PowerProvider* provider, mProviders)
     {
-        if (provider->canAction(action) &&
-            provider->doAction(action)
-           )
+        if (provider->canAction(action))
         {
-            return true;
+            if (action == PowerSuspend || action == PowerHibernate)
+            {
+                mScreenSaver.lockScreen();
+                mLoop.exec();
+            }
+            if (provider->doAction(action))
+            {
+                return true;
+            }
+            return false;
         }
     }
     return false;
--- a/lxqtpower/lxqtpower.h
+++ b/lxqtpower/lxqtpower.h
@@ -31,6 +31,8 @@
 
 #include <QObject>
 #include <QList>
+#include <QEventLoop>
+#include "lxqtscreensaver.h"
 #include "lxqtglobals.h"
 
 namespace LxQt
@@ -101,6 +103,8 @@ public slots:
 
 private:
     QList<PowerProvider*> mProviders;
+    LxQt::ScreenSaver mScreenSaver;
+    QEventLoop mLoop;
 };
 
 } // namespace LxQt


mid-kid:
Oh, also, have [the thread](https://github.com/lxde/lxqt/issues/775) that suggested I make a pull.
Please tell me if there's anything wrong with the patch, and I'll fix it, or discuss about it, or something.

lxqtpower/lxqtpower.cpp
```diff
@@ -66,11 +68,18 @@ bool Power::doAction(Power::Action action)
...
-           )
+        if (provider->canAction(action))
         {
-            return true;
+            if (action == PowerSuspend || action == PowerHibernate)
+            {
+                mScreenSaver.lockScreen();
+                mLoop.exec();
+            }
+            if (provider->doAction(action))
```
paulolieuthier: What if we get false here? `mLoop` shouldn't be running.

paulolieuthier: @mid-kid?

mid-kid: Sorry for the late reaction.
Should mLoop run after doAction? I thought it was blocking.

paulolieuthier: I mean that the provider may not be able to hibernate, for example. And so the screen shouldn't be locked.

paulolieuthier: But I see now the `canAction` is supposed to say that already.

paulolieuthier:
In my system, when it wakes up after a suspend, it shows the windows for a second before locking the screen. @mid-kid can you reproduce it as well? I'd nice if it locked the screen faster.

mid-kid:
I use i3lock, and I can't reproduce that problem. I think that depends on how long it takes to lock/how the locker handles locking.

paulolieuthier:
Merged in bdccf0641fdd7751c778a5937f3b87dd1cc17346. Thanks!

paulolieuthier:
For future reference: this is not the Right Way :copyright:. The right way to do it is described here: http://blog.martin-graesslin.com/blog/2015/01/locking-the-screen-before-system-suspends/.

surlykke:
This is not working for me: I don't have xscreensaver installed, and so when I close my laptop lid, all that happens is that a dialog comes up telling me that locking failed because xscreensaver isn't there. Of course I don't see the popup before opening the lid again, and when I dismiss it, then my laptop suspends :-) 
So I think:
- If the locking fails - suspending should proceed anyway.
- It would be nice to be able to turn off this locking feature all together. I find it quite annoying.

paulolieuthier:
@surlykke indeed, you are right. But this is the PR. I think we should move the discussion to the original issue.

mid-kid:
>Neither do I know for sure if this behaviour is acutally expected, or should be configureable or something.
Don't tell me I didn't warn you... (sorry if this sounds dicky, but I'd have liked more input before it got merged)
In either case, this discussion should be furthered in the original issue. Closing this.

paulolieuthier:
@surlykke problem fixed in lxde/liblxqt@e3f610a73b8ea162bb892be426ea087f0f07f163.

surlykke:
Not for me, I'm afraid. Behavior is the same as I described previously.

(What issue is tracking this ?)

br. Chr.

2015-09-30 18:07 GMT+02:00 Paulo Lieuthier <notifications@github.com>:

paulolieuthier:
> If the locking fails - suspending should proceed anyway.

This is what I mean was fixed. Are you sure it was not?

> (What issue is tracking this ?)

This is the one: lxde/lxqt#775.

surlykke:
2015-09-30 22:46 GMT+02:00 Paulo Lieuthier <notifications@github.com>:

> If the locking fails - suspending should proceed anyway.
>
> This is what I mean was fixed. Are you sure it was not?
>
> (What issue is tracking this ?)
>
> This is the one: lxde/lxqt#775 <https://github.com/lxde/lxqt/issues/775>.
>
>
Ok. I'm officially stupid. I was assuming you had pushed the commit to
master, so I just did a pull and tested. Looking closer at what you wrote,
you mention a commit:

e3f610a73b8ea162bb892be426ea087f0f07f163 "Warn missing xscreensaver in a
non-blocking way"

but I can't find that commit in master anywhere. I  *can* find this commit :

d036f7aa632d92822c348b852baf9b1af25b3299 - with the same commit message:  "Warn
missing xscreensaver in a non-blocking way".

The two commits differ (importantly, I think) in that in e3f610a you call
box.open() and in d036f7 you call box.exec().

Am I missing something?

br. Chr.

PR #59, opened 2015-09-01
licensefix for lxqttranslator.cpp

agaida:
@SokoloffA , @luis-pereira  - ok with the change? Copied the license from the header


paulolieuthier:
GTM.

PR #64, opened 2015-10-02
Screensaver: rely on xdg-util's xscreensaver-command

paulolieuthier:
`xdg-screensaver` works but doesn't give valuable output. We can't know if the screen is already locked or the locking failed. With `xscreensaver-command`, we have refined output, although in string form.

This has a drawback: if the user is running an app that relies on liblxqt's Screensaver (e.g. lxqt-powermanagement, lxqt-panel, lxqt-runner) outside of LXQt, the specific DE's screensaver will be triggered, but X's default. It's probably not a current common use case. Anyway, in the near future we will provide our own implementation of `org.freedesktop.ScreenSaver` in DBus and that will be solved.

Related to lxde/lxqt#833.

### DIFF
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -25,22 +25,25 @@
  *
  * END_COMMON_COPYRIGHT_HEADER */
 
-#include <QProcess>
 #include "lxqtscreensaver.h"
 #include "lxqttranslator.h"
 
+#include <QDebug>
 #include <XdgIcon>
-#include <QMessageBox>
-#include <QAction>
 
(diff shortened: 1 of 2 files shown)


paulolieuthier:
Actually, about the DBus method, there is [unfinished work](https://github.com/lxde/lxqt-powermanagement/blob/master/src/idlenesswatcher.cpp#L218) in lxqt-powermanagement that should be finished.

jleclanche:
I'm -1 on this overall. Like you said the correct fix is to provide our own ScreenSaver dbus implementation, and trigger *that* directly instead.

PR #65, opened 2015-10-16
config dialog: Protect against null pages add.

luis-pereira:
Make it resilient to null pointers pages.


palinek:
GTM

jleclanche:
gtm

luis-pereira:
@palinek The `Q_ASSERT` is removed on release builds as @jleclanche pointed out. It's there to make it stop and print a warning message containing the source code file name and line number, so a developer can debug the issue.

palinek:
I wasn't saying that it will harm in production... I was just trying to say that the check is performed twice in if one calls the QString overload...

luis-pereira:
@palinek I wasn't undertanding what you were saying. This morning I saw what you meant.
Updated PR.

palinek:
GTM :smile:

PR #67, opened 2015-10-23
Layout updates

palinek:
Support for visualizated item moving.

Maybe we should provide `LXQt::ItemMoveAnimation` as a standalone class as it can be usefull for LXQt components (the logic is stolen from lxqt-panel anyway).


jleclanche:
What is it for?

paulolieuthier:
The taskbar reordering visual feedback: lxde/lxqt-panel#253.

PR #68, opened 2015-10-26
gridlayout: Use linear animation

palinek:
We should be "light" & the QEasingCurve::Linear should be the less resources expensive one.


palinek:
...merging (agreed on IRC, that this needs users feedback)

PR #69, opened 2015-10-29
Create liblxqt_hr.ts

zvacet:


pmattern:
PR is suffering from some of the problems discussed in lxde/compton-conf#22 which should better get fixed.

PR #72, opened 2015-11-16
application: Correct signal handling & clean-up

palinek:
Correctly dispose all resources that we acquired -> close also the socketpair file descriptors.

### DIFF
--- a/lxqtapplication.cpp
+++ b/lxqtapplication.cpp
@@ -136,43 +136,69 @@ void Application::updateTheme()
 
 namespace
 {
-    int signal_sock[2];
-
-
-    void signalHandler(int signo)
+    class SignalHandler
     {
-        int ret = write(signal_sock[0], &signo, sizeof (int));
-        if (sizeof (int) != ret)
-            qCritical() << QStringLiteral("unable to write into socketpair, %1").arg(strerror(errno));
-    }
+    public:
+        static void signalHandler(int signo)
+        {
+            int ret = write(instance->mSignalSock[0], &signo, sizeof (int));
+            if (sizeof (int) != ret)
+                qCritical() << QStringLiteral("unable to write into socketpair, %1").arg(strerror(errno));
+        } 
+
+    public:
+        template <class Lambda>
+        SignalHandler(Lambda signalEmitter)
+            : mSignalSock{-1, -1}
+        {
+            if (0 != socketpair(AF_UNIX, SOCK_STREAM, 0, mSignalSock))
+            {
+                qCritical() << QStringLiteral("unable to create socketpair for correct signal handling: %1)").arg(strerror(errno));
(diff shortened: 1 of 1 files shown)


luis-pereira:
@palinek GTM

palinek:
Actually... to be 100% correct,  we need the SignalHandler object to be a (non-static) member of Application <- to avoid accessing of possibly invalid Application object in lambda emitting the signal.

@luis-pereira how do we deal with liblxqt ABI compatibility? Can I change the Application (add some private...)?

agaida:
@palinek: you can add as much stuff as you want to . thats abi compatible in the sense that the "old" abi remains unchanged (and unchanged). new symbols will only affect applications which use the new interfaces. if one want to have it nice - one should talk to the kde or Qt people how they version their symbols. i would suggest a patch version at least, maybe a new minor.

in that case one could define lxqt-abi-0-10-0, lxqt-abi-0-10-1 and so on and mark the respective symbols that way.

BTW: if you want to be (in)famous - change existing interfaces, you will receive big love from all packagers from all distributions :D

agaida:
some other sample from qt:
one can use more than one symbols layout - like this (1 marks the first alternative layout) 
<pre>
#SymbolsHelper-Confirmed: 5.4.2 amd64
libQt5Sql.so.5 libqt5sql5 #MINVER#
| libqt5sql5 #MINVER#, qtbase-abi-5-5-1
  _ZTV16QSqlDriverPlugin@Base 5.0.2
  _ZTV17QSqlResultPrivate@Base 5.1.0 1
</pre>

palinek:
@agaida the change, that I need to do, isn't about new symbols (functions). I need to add a private data member to object -> that means no API change, but the ABI will break because of different object(s) size. 

This (if used by applications compiled against the "old" liblxqt) will lead to memory corruption and SEGFAULTs...

paulolieuthier:
The correct solution here, as I'm sure you know, is to use a private class for the member data.

palinek:
> The correct solution here, as I'm sure you know, is to use a private class for the member data.

Yes, but this change will break binary compatibility this time as well.

paulolieuthier:
Yeah, but breaking once not to break anymore is alright. Packagers will have to recompile all packages. @agaida is there a better way to solve that?

agaida:
to be true i really have no clue and must read and learn a lot ...

palinek:
We need to change the libtool versioning as discussed in lxde/lxqt#488...

agaida:
@palinek Any news?

palinek:
> Any news?

I would stick with the "simple proposal" (https://github.com/lxde/lxqt/issues/488#issuecomment-215339325), but I'm not sure about opinions of others...

agaida:
GTM

PR #77, opened 2016-01-08
settings: Upgrade emitting of "changed" signals

palinek:
- Same signal settingsChanged() was emitted upon object based change (setValue(), erase()...) and
upon change made by external entity (watching the file changes) => we added other two signals
settingsChangedFromExternal() & settingsChangedByApp() to make these changes distinguishable. The signal settingsChanged() was left for backwards compatibility.

- Emitting of settingsChanged*() signals was delayed to be able to distinguish if change made on file was based upon our write/sync or by external application.

This is needed by lxde/lxqt-panel#294

### DIFF
--- a/lxqtsettings.cpp
+++ b/lxqtsettings.cpp
@@ -45,7 +45,9 @@ class LXQt::SettingsPrivate
 {
 public:
     SettingsPrivate(Settings* parent):
-        mChangeTimer(0),
+        mFileChangeTimer(0),
+        mAppChangeTimer(0),
+        mAddWatchTimer(0),
         mParent(parent)
     {
     }
@@ -53,7 +55,9 @@ public:
     QString localizedKey(const QString& key) const;
 
     QFileSystemWatcher mWatcher;
-    int mChangeTimer;
+    int mFileChangeTimer;
+    int mAppChangeTimer;
+    int mAddWatchTimer;
 
 private:
     Settings* mParent;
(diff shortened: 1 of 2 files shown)


lxqtsettings.cpp
```diff
@@ -214,7 +246,10 @@ void Settings::_fileChanged(QString path)
     // Luckily, I found a workaround: If the file path no longer exists
     // in the watcher's files(), this file is deleted.
     if(!d_ptr->mWatcher.files().contains(path))
-        d_ptr->mWatcher.addPath(path);
+        // in some situations adding fails because of non-existing file (e.g. editting file in external program)
+        if (!d_ptr->mWatcher.addPath(path) && 0 == d_ptr->mAddWatchTimer)
+            d_ptr->mAddWatchTimer = startTimer(100);
```
paulolieuthier: It seems to me we may get into a infinte loop here if the file gets deleted. Adding the file won't be work, so the timer will be started, everything again. Tell me I'm wrong.

palinek: Yes. And that is what we want.... sooner or later the file will be created (it is the settings file) and we want to know/observe it.

OK... the timeout may be greater... what do you suggest? 1 second?

paulolieuthier: OK. 1 second sounds better.

palinek: Changed...

paulolieuthier:
GTM.

palinek:
@paulolieuthier are we going to wait for someone else to review?

paulolieuthier:
It's been already 20 days, right? We've both tested it, so I think it's safe to merge it. Go ahead.

PR #78, opened 2016-01-12
[wip] translations: Add support for external translations

palinek:
Translations now can be pulled from external git repository, if enabled by LXQT_PULL_TRANSLATIONS.

Pulling of external translations files is performed in cmake phase (not in build time) into project source dir =>
- not (re)pulled in build time (after make invocation)
- translations dir is not "cleaned" in make clean
- translations dir will be "cleaned" in cmake phase if LXQT_CLEAN_TRANSLATIONS is set

TODO:
- [x] ~~unify for \*.ts \& \*.desktop translations~~ (not needed)
- [x] add support for defining the pulled directory (now it probably will not work for panel plugins etc.)
- [x] add support for selecting branch/tag/commit of translations repo to be pulled

@jleclanche @luis-pereira @paulolieuthier @PCMan Have a look on this proposal.
Is it worth continuing by this approach?

### DIFF
--- a/cmake/modules/LXQtTranslateTs.cmake
+++ b/cmake/modules/LXQtTranslateTs.cmake
@@ -168,14 +168,19 @@ function(lxqt_translate_ts qmFiles)
         endif ()
         if (NOT EXISTS "${TR_TRANSLATION_DIR}/${TR_REPO_SUBDIR}")
             message(STATUS "Setting git repository in the translations dir '${TR_TRANSLATION_DIR}' ...")
+            if (EXISTS "${TR_TRANSLATION_DIR}/.git")
+                execute_process(COMMAND rm -Rf .git
+                    WORKING_DIRECTORY  "${TR_TRANSLATION_DIR}"
+                    RESULT_VARIABLE ex_result
+                    )
 
-            execute_process(COMMAND rm -Rf .git
-                WORKING_DIRECTORY  "${TR_TRANSLATION_DIR}"
-                RESULT_VARIABLE ex_result
-                )
-            if (NOT "${ex_result}" EQUAL 0)
-                message(FATAL_ERROR "Initialization(cleanup) of translations dir failed!")
-            endif ()
+                if (NOT "${ex_result}" EQUAL 0)
+                    message(FATAL_ERROR "Initialization(cleanup) of translations dir failed!")
+                endif ()
+            endif()
+
+            # make sure the dir exist, otherwise git init will fail
+            file(MAKE_DIRECTORY "${TR_TRANSLATION_DIR}")
 
             execute_process(COMMAND "${GIT_EXECUTABLE}" init
                 WORKING_DIRECTORY  "${TR_TRANSLATION_DIR}"
--- a/cmake/modules/LXQtTranslateTs.cmake
+++ b/cmake/modules/LXQtTranslateTs.cmake
@@ -34,6 +34,11 @@
 #                           [TRANSLATION_DIR] translation_directory
 #                           [INSTALL_DIR] install_directory
 #                           [COMPONENT] component
+#                           [PULL_TRANSLATIONS [Yes | No]]
+#                           [CLEAN_TRANSLATIONS [Yes | No]]
+#                           [REPO_SUBDIR] repository_subdirectory
+#                           [TRANSLATIONS_REPO] remote_translation_repo
+#                           [TRANSLATIONS_REFSPEC] translations_remote_branch
 #                    )
 #     Output:
 #       qmFiles The generated compiled translations (.qm) files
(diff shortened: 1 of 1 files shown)


cmake/modules/LXQtTranslateTs.cmake
```diff
@@ -84,7 +92,66 @@ function(lxqt_translate_ts qmFiles)
         set(TR_TRANSLATION_DIR "translations")
     endif()
 
-    file(GLOB tsFiles "${TR_TRANSLATION_DIR}/${TR_TEMPLATE}_*.ts")
+    if (NOT DEFINED LXQT_TRANSLATIONS_REPO)
+        set(LXQT_TRANSLATIONS_REPO "https://github.com/palinek/translations.git")
```
f2404: Sorry, do you want the translations to be stored in your own repository (not one of lxde's)?

palinek: Of course not... it's only a testing one, because I'm lacking write permission in the newly created https://github.com/lxde/translations

jleclanche: I have fixed permissions.

palinek:
How about this, guys?

palinek:
ping
@jleclanche

jleclanche:
Gonna try to look into this this week...

palinek:
> Gonna try to look into this this week...

Any news?

paulolieuthier:
I just tested it and it works very well. My two cents:

- The translations repository should be names lxqt-translations.
- Is there a way to pull all panel and plugins translation at once?
- IMO pulling should be enabled by default, but a lack of connection (or any other problem) shouldn't trigger an error, only a CMake warning.

palinek:
> The translations repository should be names lxqt-translations.

@jleclanche didn't want to use such name -> repo `lxde/translations` will be used for both LXDE & LXQt

> Is there a way to pull all panel and plugins translation at once?

I'm not aware of any such possibility.

> IMO pulling should be enabled by default, but a lack of connection (or any other problem) shouldn't trigger an error, only a CMake warning.

This is no big deal to implement, but let's wait for opinion of others.

pmattern:
What I personally wonder with respect to this approach is its eventual impact on the availability of previews of objects to be translated while doing the actual translation work.
Those previews of Qt Linguist would sure stop working if the TS files are out of the source tree, wouldn't they? I do not know whether tools like Pootle or Transifex, the support of which is the actual aim of the changes discussed here if I'm not mistaken, do feature something similar in the first place.
But considering my personal experience I'd say that missing previews like that would sure be quite a loss.

palinek:
> Those previews of Qt Linguist would sure stop working if the TS files are out of the source tree, wouldn't they?

I anyone would wish to do the translation in the old way he/she can do it the similar way as today -> checkout translation of particular component (done by cmake), make the translation with linguist, commit, push (, make PR)....

jleclanche:
BTW, I just realized I didn't say this here but the Pootle team confirmed they are releasing PootleFS some time in March. I'm super hopeful we can get that updated then.

luis-pereira:
What am I supposed to do here ?

jleclanche:
@luis-pereira @lxde/lxqt Update on the plan. Last time we talked about it, we decided not to wait on the Pootle team anymore and carry on with the original idea to move translations to a single lxde/translations repository but keep the pull request model *for now*. It'll keep the history clean and still make it easier for translators to contribute.

So for now, see if we can get that model working before the release next month.

paulolieuthier:
@luis-pereira: @palinek asked for a review, and I think you're the build system master.

luis-pereira:
@palinek It has conflicts. May I rebase and push ?

palinek:
> @palinek It has conflicts. May I rebase and push ?

Of course...

palinek:
rebased

luis-pereira:
> What I personally wonder with respect to this approach is its eventual impact on the availability of previews of objects to be translated while doing the actual translation work.

@pmattern If a translator wants the previews, it can be achieved by pulling the repo with the translations. The translations have to be updated to reflect the new paths, obviously.

palinek:
@luis-pereira There are a bit too many commit here. How about squashing them together somehow?

luis-pereira:
@palinek Go ahead.

palinek:
OK... squashed & rebased. Are we going to merge it that way?

luis-pereira:
@palinek Do whatever you find it's the best.
Two questions:

- What about .desktop files ? How will we handle them ?
- Doesn't Pootle need the lxqt-foo.ts file ?

palinek:
> What about .desktop files ? How will we handle them ?

@jleclanche decided they will be handled the same way as currently

> Doesn't Pootle need the lxqt-foo.ts file ?

I don't know.... @jleclanche ?

jleclanche:
I don't think it does.

luis-pereira:
@jleclanche Ok. I'm wondering, how can it add a new language without the .ts file ?

palinek:
> @jleclanche Ok. I'm wondering, how can it add a new language without the .ts file ?

me too :)

PR #80, opened 2016-02-02
LXQtTheme: Remove (outdated) "singleclick" qss assembly

palinek:
fixes lxde/lxqt#940

Do I miss something or is the reasoning (i've written into the commit log message) 100% valid?

### DIFF
--- a/lxqtsettings.cpp
+++ b/lxqtsettings.cpp
@@ -498,14 +498,7 @@ QString LXQtTheme::previewImage() const
  ************************************************/
 QString LXQtTheme::qss(const QString& module) const
 {
-    QString styleSheet = d->loadQss(QStringLiteral("%1/%2.qss").arg(d->mPath, module));
-
-    // Single/double click ...........................
-    Settings s("desktop");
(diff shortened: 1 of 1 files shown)


paulolieuthier:
Good catch. 100% valid AFAIK. GTM.

palinek:
> Good catch. 

...I wasn't good enough to get it straight and had to dig deeply into the QT core :disappointed:

paulolieuthier:
Good to have you on board. :smile:

pmattern:
@palinek 
83c93de seems to fix the problem depicted in lxde/lxqt#940.
Interestingly, lxqt-panel did not have to be recompiled against 83c93de but the problem ceased once the new liblxqt version was installed (and did not come up again after recompiling the panel against 83c93de).

palinek:
> Interestingly, lxqt-panel did not have to be recompiled against 83c93de

It's not a surprise, the change does not break API nor ABI compatibility... so no need to recompile any other component using it.

palinek:
...merging

PR #81, opened 2016-02-08
SingleApplication: Use QCoreApplication::exit() instead of ::exit()

palinek:
fixes lxde/lxqt#951

But I'm not sure if this is the right way...the application's `exec()` will return "immediately", but all the initialization (outside of QApplication/.../SingleApplication) will be performed (creating widgets etc.).

### DIFF
--- a/lxqtsingleapplication.cpp
+++ b/lxqtsingleapplication.cpp
@@ -32,6 +32,7 @@
 #include <QDBusMessage>
 #include <QWidget>
 #include <QDebug>
+#include <QTimer>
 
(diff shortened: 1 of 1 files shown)


npmiller:
I actually think that this is much better this way since it allows to do some other work before exiting. Typically that could be parameter parsing, for example the application could have options to send dbus messages to the main instance.

Although adding a way for the inheriting class to know whether or not it will exit when hitting the `exec`, whether or not it is the first instance would be valuable. It would allow inheriting classes to skip all the initialization.

PR #83, opened 2016-03-04
A few size corrections.

tsujan:


configdialog/lxqtpageselectwidget.cpp
```diff
@@ -102,12 +103,14 @@ QSize PageSelectWidget::sizeHint() const
     int w = 0;
     for(int i=0; i< count(); ++i)
     {
-        QRect rect = fontMetrics().boundingRect(QRect(), Qt::AlignLeft | Qt::TextWordWrap, item(i)->text());
+        QRect rect = fontMetrics().boundingRect(QRect(), Qt::AlignHCenter | Qt::TextWordWrap, item(i)->text());
         w = qMax(w, rect.width());
     }
+    // consider the icon size too
+    w = qMax(w, QApplication::style()->pixelMetric(QStyle::PM_IconViewIconSize));
```
palinek: Just curious... why are you using the QApplication::style() ?

tsujan: > why are you using the QApplication::style() ?

Although the view icon size is set to 32 px in Fusion, it can have different values in other styles and some styles provide users with a setting for it.

palinek: hm.... I was asking about `QApplication::style()` vs. `this->style()`

tsujan: > I was asking about...

Theoretically there shouldn't be any difference because there's no hard-coded value here. I just don't trust hard-coded styles because of my experience from Kvantum: there are so many good devs who make strange mistakes when it comes to QStyle.

paulolieuthier:
GTM.

configdialog/lxqtpageselectwidget.cpp
palinek: But no matter if the `QApplication::style()` or `this->style()` is used, by not providing the widget to `pixelMetric()` call you are denying styling for this particular widget via StyleSheet. Why?

tsujan: Please see `qcommonstyle.cpp` → `QCommonStyle::pixelMetric` → `case PM_ScrollBarExtent:`! `widget` has no part in it.

palinek: > Please see qcommonstyle.cpp → QCommonStyle::pixelMetric → case PM_ScrollBarExtent:! widget has no part in it.

That's true, but `QCommonStyle` is the "base" style. See the `QStyleSheetStyle::pixelMetric()`...

tsujan: Do you see any trace of `QWidget *w` in `qstylesheetstyle.cpp` → `QStyleSheetStyle::pixelMetric` → `case PM_ScrollBarExtent:`? It's just in `QRenderRule rule = renderRule(w, opt);` before `switch`.

Anyway, if you worry that something may be lost, add it -- it'll neither make any difference nor do any harm, AFAIK.

palinek: > It's just in QRenderRule rule = renderRule(w, opt); before switch.

Of course... the rule is taken based on the widget. If the widget is not present (nullptr), the specific rule (if existing) will not be chosen/found.

tsujan: OK, `rule.size()` is used. So it may make a difference. Please add it yourself!

tsujan: ... And if you want, change QApplication::style() to style().

palinek: Opened PR #88 ...

tsujan: Great!

PR #84, opened 2016-04-08
LXQtCompilerSettings: Add QT_USE_QSTRINGBUILDER to the compile defini…

luis-pereira:
…tions

It makes the '+' will automatically be performed as the QStringBuilder '%'
everywhere.

Disclaimer: This is not entirely source compatible.


palinek:
I'm OK with this (but didn't test the change :smile: )

PR #86, opened 2016-04-21
Icon theme fallback

luis-pereira:


jleclanche:
gtm

luis-pereira:
Added `const` qualifier.

paulolieuthier:
GTM.

PR #87, opened 2016-04-22
Improve iterator performance

luis-pereira:


lxqtsettings.cpp
```diff
@@ -631,10 +631,12 @@ SettingsCache::SettingsCache(QSettings *settings) :
...
-   foreach (QString key, mSettings.allKeys())
-   {
-       mCache.insert(key, mSettings.value(key));
-   }
+    const QStringList keys = mSettings.allKeys();
+
+    const int N = keys.size();
+    for (int i = 0; i < N; ++i) {
+        mCache.insert(keys.at(i), mSettings.value(keys.at(i)));
+    }
```
paulolieuthier: How is this faster than the regular range-based for with const reference, which is more readable?

luis-pereira: It automatically takes a copy of the container when it enters a foreach loop.

paulolieuthier: I see, you're right. Seeing the implementation of `allKeys()`, I say it's quite expensive.

paulolieuthier:
Ship it.

PR #89, opened 2016-04-27
Lp ext translations

luis-pereira:


jleclanche:
GTM

cmake/modules/LXQtTranslateTs.cmake
```diff
@@ -168,14 +168,16 @@ function(lxqt_translate_ts qmFiles)
         endif ()
         if (NOT EXISTS "${TR_TRANSLATION_DIR}/${TR_REPO_SUBDIR}")
             message(STATUS "Setting git repository in the translations dir '${TR_TRANSLATION_DIR}' ...")
+            if (EXISTS "${TR_TRANSLATION_DIR}/.git")
```
palinek: This is needles, because with -f parameter the rm will not fail if the directory doesn't exist ... OK not needles, but not a "real optimization".

palinek:
Why a new PR based on #78, if that wasn't merged yet?

cmake/modules/LXQtTranslateTs.cmake
luis-pereira: @palinek It's not an optimization. It actually fails when pulling `lxqt-panel/panel` translations.
See it  [here](https://gist.github.com/luis-pereira/7417b15d893bd1203fe64e19b08c8905)

luis-pereira:
@palinek Not sure if I understood the meaning of your question. Just placed a PR to make it visible.

palinek:
> Not sure if I understood the meaning of your question. Just placed a PR to make it visible.

I meant.... why didn't you push your commits directly into #78?

luis-pereira:
That's your branch. It's up to you what gets in.

PR #92, opened 2016-05-20
configdialog: Update icons in ThemeChanged event

palinek:
We don't need/want to handle global settings change. lxqt-qtplugin is handling that.
We just need to know that the theme was changed.


luis-pereira:
GTM.
That code is from the pre-plugin era.

PR #94, opened 2016-05-31
SingleApplication: Only create an DBus adaptor when needed

luis-pereira:
Save an allocation and helps static analyzers.


palinek:
GTM

paulolieuthier:
GTM.

PR #98, opened 2016-06-30
lxqtscreensaver: Makes it a opaque pointer class

luis-pereira:
It helps with where the implementation details of a library may be hidden
from its users and changes to the implementation can be made to a library
without breaking binary compatibility.

### DIFF
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -32,40 +32,32 @@
 #include <XdgIcon>
 #include <QMessageBox>
 #include <QAction>
+#include <QPointer>
+#include <QProcess>
+#include <QCoreApplication> // for Q_DECLARE_TR_FUNCTIONS
 
-using namespace LXQt;
+namespace LXQt {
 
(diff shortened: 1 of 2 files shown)


lxqtscreensaver.cpp
```diff
@@ -32,40 +32,32 @@
...
-    return ret;
-}
-
-void ScreenSaver::lockScreen()
-{
-    m_xdgProcess->start("xdg-screensaver", QStringList() << "lock");
-}
-
-void ScreenSaver::xdgProcess_finished(int err, QProcess::ExitStatus status)
+void ScreenSaverPrivate::_l_xdgProcess_finished(int err, QProcess::ExitStatus status)
```
paulolieuthier: What's the reason for the `_l_` prefix?

paulolieuthier:
GTM.

luis-pereira:
Kde uses  k
Qt uses q
lxqt ..... l

agaida:
@luis-pereira hey - hit the merge button please :P

PR #99, opened 2016-07-01
Don't lock the screen when it's already locked.

luis-pereira:
The status of the screensaver wasn't taken into account when locking it.
Now we only lock it if it's not already locked. This way we don't get the
annoying error message stating that there was an error starting the
screensaver.

Part of the code is taken from the Chromium project. Adapted to fit our
purposes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTOUIC ON)
 
+find_package(X11 REQUIRED QUIET)
 find_package(Qt5Widgets REQUIRED QUIET)
 find_package(Qt5DBus REQUIRED QUIET)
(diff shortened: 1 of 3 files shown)


luis-pereira:
@paulolieuthier Test in conjunction with https://github.com/lxde/liblxqt/pull/98

agaida:
merged both PRs and build  local - works nearly perfect. Please don't throw stone on me, but:
* the wakeup after first suspend or hibernate is ok and as expected, screen is locked without any delay
* second and more - screen will be locked after wakeup - better than before, but ...

paulolieuthier:
Although it makes me sad that we'll be adding X11 stuff back to liblxqt, GTM. Works well (@agaida: that's another issue).

agaida:
adding X11 stuff is the right thing™ for the time being - give me a working and stable Wayland full supported right now and i will change my mind in just a second :)

luis-pereira:
@paulolieuthier I'm convinced that we can do this without depend here on X11. But for the time being, this is the possible solution.

paulolieuthier:
Great.

PR #101, opened 2016-08-03
Fix lupdate warning missing Q_OBJECT macro

luis-pereira:
We don't need to add the Q_OBJECT macro. Just use the
Q_DECLARE_TR_FUNCTIONS macro.


tsujan:
GTM.

paulolieuthier:
GTM.

PR #102, opened 2016-08-04
Update DBus Notify method annotation name

luis-pereira:
Just following moc suggestion. And making a warning go away.


paulolieuthier:
GTM.

agaida:
GTM

PR #104, opened 2016-08-23
Add XDG_CONFIG_DIRS support

palinek:
ref lxde/lxqt#349, lxde/lxqt#385, lxde/lxqt#250

I don't know if this is enough to close the issues (each component needs to be checked if it doesn't use "plain" QSettings for its configuration).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ list(APPEND CMAKE_MODULE_PATH
 )
 
 option(UPDATE_TRANSLATIONS "Update source translation translations/*.ts files" OFF)
+option(WITH_XDG_DIRS_FALLBACK "Use our XDG_CONFIG_DIRS fallback" ON)
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
@@ -324,6 +325,12 @@ target_compile_definitions(${LXQT_LIBRARY_NAME}
     PRIVATE "$<$<CONFIG:Release>:QT_NO_DEBUG_OUTPUT>"
     PRIVATE "$<$<CONFIG:Release>:QT_NO_WARNING_OUTPUT>"
 )
+
+if (WITH_XDG_DIRS_FALLBACK)
+    target_compile_definitions(${LXQT_LIBRARY_NAME}
+        PRIVATE "WITH_XDG_DIRS_FALLBACK"
+    )
+endif ()
 
 target_include_directories(${LXQT_LIBRARY_NAME}
     INTERFACE "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${LXQT_LIBRARY_NAME}>"
--- a/lxqtsettings.cpp
+++ b/lxqtsettings.cpp
@@ -43,7 +43,7 @@ using namespace LXQt;
 class LXQt::SettingsPrivate
 {
 public:
-    SettingsPrivate(Settings* parent):
+    SettingsPrivate(Settings* parent, bool useXdgFallback):
         mFileChangeTimer(0),
         mAppChangeTimer(0),
         mAddWatchTimer(0),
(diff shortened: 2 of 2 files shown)


gilir:
Great, thanks for time on this issue :-)

If I understand right, you implement it with a build flag, I'm not sure it's practical for a downstream point of view. Example : if you want to build a custom distro Debian-based, you have to rebuild the liblxqt package just to set some custom settings. Usually, you don't have to modify the way packages are build just to change some settings.

On Lubuntu, we heavily use the XDG_CONFIG_DIRS with LXDE. We just have to add /etc/xdg/Lubuntu in the beginning of XDG_CONFIG_DIRS, putting all the modified files here, and it's done. XDG_CONFIG_DIRS is read at runtime, from the beginning of the list of directories in XDG_CONFIG_DIRS. It's also possible for any people to build a custom distro just by adding /etc/xdg/another_diretory to XDG_CONFIG_DIRS, and put their settings into this.

palinek:
> If I understand right, you implement it with a build flag, I'm not sure it's practical for a downstream point of view. Example : if you want to build a custom distro Debian-based, you have to rebuild the liblxqt package just to set some custom settings.

It's default on -> https://github.com/lxde/liblxqt/pull/104/commits/b220d13538b52afdbce7da00fefa4a5796d1fe53#diff-af3b638bc2a3e6c650974192a53c7291R127

I've added the option to easily turn it of in case someone needs that.

gilir:
Sorry, I didn't read the code correctly the first time I though the build flag set only 1 custom dir ... Just reading it again, and I think it does what I had in mind. Thanks :-)
I'll test it later this week to confirm this.

palinek:
I've made the proper implementation directly into qt -> https://github.com/qt/qtbase/compare/dev...palinek:xdg_fallback

Can someone review & test for correctness before submitting it to Qt?

palinek:
Just a note... this will not be needed if the patch(https://codereview.qt-project.org/#/c/169245/) will be accepted in Qt.

paulolieuthier:
From the [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html): 

> "The order of base directories denotes their importance; the first directory listed is the most important. When the same information is defined in multiple places the information defined relative to the more important base directory takes precedent."

It makes me think we shouldn't just search for the first available settings file in the directory found. Every setting information should be searched until found. For example:

```bash
$ echo $XDG_CONFIG_DIRS
/etc/xdg/Lubuntu:/etc/xdg
$ cat /etc/xdg/Lubuntu/lxqt/lxqt.conf
[General]
icon_theme=oxygen
$ cat /etc/xdg/lxqt/lxqt.conf
[General]
icon_theme=elementary
theme=light
```

So `icon_theme` should be "oxygen", because of precedence of `/etc/xdg/Lubuntu`, but `theme` should be "light" even though it's not defined in `/etc/xdg`.

palinek:
> It makes me think we shouldn't just search for the first available settings file in the directory found. Every setting information should be searched until found.

Why are you thinking, we are searching only the first one? The logic wants to be "if user specific config file is empty -> search for all system-wide configs and put all their keys/values into the user specific one (so if multiple configs found, the values will be overwritten)".
Note: that this is a workaround, not a proper solution.

palinek:
Hm... but now I've realized that we should iterate the directories in XDG_CONFIG_DIRS in reverted order to achieve correct prioritization.

paulolieuthier:
> Why are you thinking, we are searching only the first one?

Because we're [stopping in the first fle found](https://github.com/lxde/liblxqt/pull/104/files#diff-ca66c268a6ce420d8e5e20768edccca5R77), aren't we?

palinek:
You're right ... pushed the fixup (also for the priority).

paulolieuthier:
Great. The only thing I can think of now is that maybe we should monitor all those files for modifications, and read them all again (maybe this time in order, only looking for diffs).

paulolieuthier:
Fair enough. GTM.

PR #111, opened 2016-11-16
build: Enable/disable WITH_XDG_DIRS_FALLBACK based on Qt version

palinek:
Use homemade Xdg dirs settings fallback by default if Qt < 5.9.0 and
force disabling the workadound for Qt >= 5.9.0 (as from 5.9.0 the
proper XDG_CONFIG_DIRS support is directly in QSettings).

ref. lxde/lxqt#1158

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,7 +275,16 @@ target_compile_definitions(${LXQT_LIBRARY_NAME}
     PRIVATE "$<$<CONFIG:Release>:QT_NO_WARNING_OUTPUT>"
 )
 
+if (Qt5Core_VERSION VERSION_LESS "5.9.0")
+    if (NOT DEFINED WITH_XDG_DIRS_FALLBACK)
+        set(WITH_XDG_DIRS_FALLBACK ON)
+    endif ()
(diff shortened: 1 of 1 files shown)


tsimonq2:
In Debian and Ubuntu, the proper XDG_CONFIG_DIRS support has been backported to the Qt 5.7.1 packages, so when that gets released and in the Ubuntu archive (already in Debian Sid), this workaround will not be needed.

So if this gets merged, either me or @gilir will have to make a patch reverting this (if LXQt 0.12 lands before Qt 5.9), so we can use the built in Qt functionality.

(at least **I** would prefer that over any homemade workaround)

palinek:
> So if this gets merged, either me or @gilir will have to make a patch reverting this (if LXQt 0.12 lands before Qt 5.9), so we can use the built in Qt functionality.

You don't need to revert it, you just need to update your build scripts to use `-DWITH_XDG_DIRS_FALLBACK=off` with cmake

tsimonq2:
Oh ok, I misunderstood, sorry.

agaida:
GTM

PR #115, opened 2017-01-26
Fix warnings

luis-pereira:


luis-pereira:
@palinek Just used the Qt mnemonic.

palinek:
merged into master...

PR #117, opened 2017-03-31
lxqtapplication: Avoid wrong signal handling

palinek:
Add the application object as receiver to QObject::connect() -> to
remove the connection automaticaly when the application object is
destructed.
W/o the automatic connection removal we could end in SEGFAULT if UNIX
signal is delivered after the QApplication destruction.

(this was a corner case with almost zero probablity, but now the UNIX
signal handling should be 100% correct)

### DIFF
--- a/lxqtapplication.cpp
+++ b/lxqtapplication.cpp
@@ -148,7 +148,7 @@ namespace
 
     public:
         template <class Lambda>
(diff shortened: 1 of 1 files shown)


palinek:
This handles the corner case discussed in #72 w/o the need of ABI break.

agaida:
lgtm

PR #118, opened 2017-04-06
Fixes

luis-pereira:


agaida:
Apply, build and run great. LGTM -- we should mention somewhere in the wiki that the minimum version for LXQt is now 5.5

lxqtsettings.cpp
```diff
@@ -80,7 +80,8 @@ class LXQt::SettingsPrivate
                     if (dir.cd(mParent->organizationName()) && dir.exists(file_name))
                     {
                         QSettings system_settings{dir.absoluteFilePath(file_name), QSettings::IniFormat};
-                        for (const QString & key : system_settings.allKeys())
+                        const QStringList keys = system_settings.allKeys();
+                        for (const QString & key : const_cast<const QStringList&>(keys))
```
palinek: The `const_cast` is unneeded here, `keys` variable is already `const`

luis-pereira: Two solutions for the same problem.

agaida:
@luis-pereira @palinek - ready?

palinek:
> - ready?
yes

PR #119, opened 2017-06-06
HtmlDelegate: Fix sizeHint() computation

palinek:
- w/o calling the QTextDocument::adjustSize() the size computation is
not correct
- the option.rect is not set if only the sizeHint() is needed by Qt

Note: this was revealed while struggling with lxde/lxqt-runner#58


agaida:
GTM

PR #120, opened 2017-06-07
Move lxqt version

agaida:


palinek:
But this is definitely not sufficient...with these changes, you only "override" the version for the `liblxqt`, but not for including into other components. Other components will still see/use the version parameters from `lxqt-buildtools` -> https://github.com/lxde/liblxqt/blob/master/cmake/lxqt-config.cmake.in#L18.

agaida:
damn, i know there was something else :) - will fix it today

agaida:
@luis-pereira @palinek any objections?

palinek:
GTM

luis-pereira:
GTM.

PR #121, opened 2017-06-28
lxqtpowermanager: Do logout before reboot/shutdown

palinek:
Try to terminate the running processes/modules in prefered order (the
session's logout procedure should know how to do it). Aftert that invoke
the particular reboot/shutdown.

W/o the logout before reboot/shutdown the reboot/shutdown provider
(namely systemd) was terminating all proceses in undefined (random?)
order which could lead to insane shutdown and/or deadlocks with poorly
written applications/libraries.

closes lxde/lxqt#1311

### DIFF
--- a/lxqtpowermanager.cpp
+++ b/lxqtpowermanager.cpp
@@ -177,6 +177,7 @@ void PowerManager::reboot()
         MessageBox::question(tr("LXQt Session Reboot"),
                              tr("Do you want to really restart your computer? All unsaved work will be lost...")))
     {
+        m_power->logout();
         m_power->reboot();
     }
 }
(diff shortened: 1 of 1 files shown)


agaida:
GTM

palinek:
Nope... this isn't a good way:
1. the shutdown process is triggered immediately after triggering the session's logout -> so both processes are ongoing simultaneously -> so this can't solve any problem
2. even when I tried to make the `logout()` want for DBus response (wait to finish the session's logout), it isn't OK. Because after the `logout` process is finished, we're getting error while calling the `shutdown` to systemd (some kind of "unauthorized" error....I didn't investigate it deeper yet).

agaida:
@palinek - so we should drop this approach and handle it in the session i guess?

PR #123, opened 2017-07-07
lxqtpower: Allow disabling of lxqt-session provider

palinek:
To allow usage of LXQt::Power object inside of lxqt-session (handling
the called method of LXQtSessionProvider) we need to avoid recurently
calling lxqt-session itself.

supersedes the #121 

ref. lxde/lxqt#1311


agaida:
apply and build fine - GTM

PR #131, opened 2018-03-18
Back light control added.

selairi:
A class to control backlight in laptops and other devices has been added.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,8 @@ set(LXQT_VERSION ${LXQT_MAJOR_VERSION}.${LXQT_MINOR_VERSION}.${LXQT_PATCH_VERSIO
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs) # Standard directories for installation
 
+add_subdirectory(lxqtbacklight/linux_backend/driver)
+
 #-----------------------------------------------------------------------------
 # Release is the default build type
 #-----------------------------------------------------------------------------
@@ -54,6 +56,9 @@ set(PUB_HDRS
     lxqtgridlayout.h
     lxqtrotatedwidget.h
     lxqtglobals.h
+    lxqtbacklight.h
+    #lxqtbacklight/virtual_backend.h
+    #lxqtbacklight/linux_backend/linuxbackend.h
 )
 
 set(PUBLIC_CLASSES
@@ -76,6 +81,7 @@ set(PUBLIC_CLASSES
     GridLayout
     RotatedWidget
     Globals
+    Backlight
 )
 
 set(SRCS
@@ -98,6 +104,9 @@ set(SRCS
     lxqtnotification.cpp
     lxqtgridlayout.cpp
     lxqtrotatedwidget.cpp
+    lxqtbacklight.cpp
+    lxqtbacklight/virtual_backend.cpp
+    lxqtbacklight/linux_backend/linuxbackend.cpp
 )
 
 set(MOCS
@@ -106,6 +115,7 @@ set(MOCS
     lxqtsettings.h
     lxqtscreensaver.h
     lxqtapplication.h
+    lxqtbacklight.h
 
     configdialog/lxqtconfigdialog.h
     configdialog/lxqtpageselectwidget.h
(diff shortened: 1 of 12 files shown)


tsujan:
Shouldn't `LinuxBackend::isBacklightAvailable()` consult `actual_brightness` to know if the laptop monitor is turned off (and only an external monitor is on), so that the slider could be disabled in such cases? Or, preferably, is there another way to know about an off monitor?

I ask because showing an enabled slider with a zero value may be a little confusing when only an external monitor is working.

selairi:
> Shouldn't LinuxBackend::isBacklightAvailable() consult actual_brightness to know if the laptop monitor is turned off (and only an external monitor is on), so that the slider could be disabled in such cases? Or, preferably, is there another way to know about an off monitor?

I have just changed and now it consults bl_power too. Then it can detects if the screen is turned off.

tsujan:
> I have just changed and now it consults bl_power too.

Interesting! Here, with an external LCD and the laptop screen off, `bl_power` shows 4 (I don't know why 4) but when the laptop screen is on too, it shows 0. So, your calculation should be correct. I'll try it as soon as I find some time.

selairi:
> Here, with an external LCD and the laptop screen off, `bl_power` shows 4

In this link you can read kernel documentation:
[https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-backlight](https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-backlight)

`bl_power` sets to 0, turns on backlight.
`bl_power` sets to 4 , turns off backlight.

But I have done some test and other values, like 1, turn off the backlight.

I was thinking about Backlight::setBacklight(-1) turns off the screen. It could be used to turn off screen by LXQtPowerManagement when computer is idle for some time. What do you think?

tsujan:
> It could be used to turn off screen by LXQtPowerManagement when computer is idle for some time. What do you think?

I like the idea and think its implementation is very needed --  `xset` will be unusable sooner or later.

Some thoughts:

It shouldn't see video playing -- fullscreen or not -- as idle time. I don't know how video playing can be detected (can it?) or how video players like SMPlayer or VLC tell about their activities (I think whatever method they have may be detectable by the program in question) but if needed, we could consult their sources.

More importantly, the idle timer should be able to be disabled/enabled by a shortcut.

selairi:
> It shouldn't see video playing -- fullscreen or not -- as idle time. I don't know how video playing can be detected (can it?) or how video players like SMPlayer or VLC tell about their activities (I think whatever method they have may be detectable by the program in question) but if needed, we could consult their sources.

This is a good point. It is so good that I have been thinking for hours about it. I am going to write Backlight::setBacklight(-1) turns off the screen feature, but I am not going to suggest that LXQtPowerManager turn off the screen. Maybe, it could set backlight to lower levels in order to save battery.

And why have I changed my mind?
In X11 you can prevent the screen goes blank. But it looks like preventing screen lock is an entirely missing feature in Wayland:

[https://askubuntu.com/questions/969398/how-to-prevent-the-screen-lock-while-watching-video-netflix-youtube-amazon-pr?noredirect=1&lq=1](https://askubuntu.com/questions/969398/how-to-prevent-the-screen-lock-while-watching-video-netflix-youtube-amazon-pr?noredirect=1&lq=1)

There is a simple solution, don't blank or lock the screen when there is a window in fullscreen mode. But can I check the mode of the other windows in Wayland?

Maybe I am wrong, but it looks like Wayland is not going to define protocols for taking screenshots, checking windows lists or changing screen size and they will be written by window-compositors.

An example:

If we wanted to write a lightweight tool for taking screenshots, we should write several backends for KDE, Gnome, Weston,... or wait until Freedesktop write a protocol for it:

[https://bugs.freedesktop.org/show_bug.cgi?id=98894](https://bugs.freedesktop.org/show_bug.cgi?id=98894)

But the Freedesktop interface only opens the screenshot tool selected by the window-compositor:

[https://github.com/flatpak/xdg-desktop-portal/blob/master/data/org.freedesktop.impl.portal.Screenshot.xml](https://github.com/flatpak/xdg-desktop-portal/blob/master/data/org.freedesktop.impl.portal.Screenshot.xml) 

Wayland mode of LXQt will use KWin in the future, because we have used the tools provided by KDE. We have told about it:
[https://github.com/lxde/lxqt/issues/1011#issuecomment-337133689](https://github.com/lxde/lxqt/issues/1011#issuecomment-337133689)

Then is better to wait until LXQt select a Wayland compositor or Wayland compositors write some protocols. In my humble opinion, Wayland compositors are not going to write those protocols and LXQt won't be able to be independent of Wayland compositor .

selairi:
> I'm very glad that you joined LXQt.

You are doing a good work too.

I have just added the backlight power control to the library. I think that this interface is enough simple.

agaida:
@tsujan @selairi - ready?

tsujan:
@agaida IMHO, test it and if it works OK, merge it. I was so busy...

agaida:
GTM

selairi:
Backlight control works properly. :-D

2018-04-29 22:31 GMT+02:00 Alf Gaida <notifications@github.com>:

selairi:
I have some problems compiling lxqt-admin:

lxqt-admin/lxqt-admin-time/timeadmindialog.cpp:42:87: error: invalid use of
incomplete type ‘class LXQt::Settings’

????

2018-04-29 23:50 GMT+02:00 P L. Lucas <selairi@gmail.com>:

agaida:
erm - given the fact that it works before - we should have a look into liblxqt i guess

selairi:
lxqt-admin/lxqt-admin-time/timeadmindialog.cpp needs to add:


#include <LXQt/Settings>




2018-04-29 23:59 GMT+02:00 P L. Lucas <selairi@gmail.com>:

agaida:
could you please add it?

tsujan:
It correctly compiles here but my liblxqt is from 2018/03/18

PR #136, opened 2018-04-30
No in source builds

luis-pereira:
lxqt-build-tools version should probably be updated. And the minimum required here, updated accordingly,


agaida:
Full ACK - we are behind the schedule, but this is an important piece. Lets do it now.

PR #138, opened 2018-05-10
Normalize backlight value

tsujan:
Fixes https://github.com/lxqt/lxqt/issues/1485

Instead of filtering out values outside the range, normalize them to deal with `QSlider` correctly.


agaida:
GTM

PR #141, opened 2018-05-18
Don't find packages quietly

luis-pereira:
Stop using QUIET with find_package()


agaida:
:) - seems to be a lag in github

PR #143, opened 2018-05-21
build: Make PolkitQt5-1 optional

Chiitoo:
Would something like this be acceptable and sane?

If there is a hard dependency on polkit, I've clearly missed it.  :]

The comment and the description of the option could probably be better, but this is just something I came up with on a short notice.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,7 +160,14 @@ find_package(X11 REQUIRED)
 find_package(Qt5 ${QT_MINIMUM_VERSION} CONFIG REQUIRED Widgets DBus X11Extras LinguistTools)
 find_package(Qt5Xdg ${QTXDG_MINIMUM_VERSION} REQUIRED)
 find_package(KF5WindowSystem REQUIRED)
-find_package(PolkitQt5-1 REQUIRED)
+
+# Optionally include the PolkitQt5-1 module.
+option(BUILD_POLKIT "Install the PolkitQt5-1 files." ON)
+
+if (BUILD_POLKIT)
+    find_package(PolkitQt5-1 REQUIRED)
+endif ()
+
 message(STATUS "Building ${PROJECT_NAME} with Qt ${Qt5Core_VERSION}")
 
 QT5_ADD_DBUS_INTERFACE(DBUS_INTERFACE_SRCS
@@ -367,7 +374,9 @@ install(FILES ${LXQT_CONFIG_FILES}
     COMPONENT Runtime
 )
 
-install(FILES ${POLKIT_FILES} DESTINATION "${POLKITQT-1_POLICY_FILES_INSTALL_DIR}")
+if (PolkitQt5-1_FOUND)
+    install(FILES ${POLKIT_FILES} DESTINATION "${POLKITQT-1_POLICY_FILES_INSTALL_DIR}")
+endif ()
 
 #************************************************
 # Create and install pkgconfig file


Chiitoo:
Would it be feasible to make the 'lxqt-backlight' feature optional, then, or simply disable it if Polkit is not around?

I'll fully understand if it's not something you want to do.  :]

(I got rid of Polkit/Consolekit\etc. already back when I was using KDE 3 and 4, and I'm just not very keen on including them in my systems still.)

agaida:
for upstream i would guess a clear no - so i guess you will need to maintain your personal patchset

luis-pereira:
@Chiitoo I do agree with @agaida. That's not a viable feature for upstream. That's a too narrow use case for upstream.
As @agaida said, you can maintain your own patch. It's not a big deal since the patch is so simple. Merge/Rebase conflicts will not be frequent. And trivial to resolve.

luis-pereira:
@Chiitoo In the case you intend to maintain the patch keep in mind that part or all `lxqt-config-brightness` should also be made optional. Otherwise it will become very odd to the user.

Chiitoo:
Yeah, I realise that now.  :]

Thanks again!

jubalh:
In general I agree with @agaida and @luis-pereira. However:

> That's a too narrow use case for upstream. As @agaida said, you can maintain your own patch. It's not a big deal since the patch is so simple. Merge/Rebase conflicts will not be frequent. And trivial to resolve.

If I see it right it's only about having ifdefs around the polkit/backlight blocks.
If @Chiitoo already has patches it means we don't have to work on this. Officially we could only support building *with* polkit (and the switch will be on by default).

But it would allow Gentooers to compile it without polkit if they whish too. They wouldnt need to rebase their patches upon every release and as I see it there would be not additional work for us neither.

> In the case you intend to maintain the patch keep in mind that part or all `lxqt-config-brightness` should also be made optional.

I think in the Gentoo world people know that when they compile without certain use flags the program will lack certain functionality. So we as upstream could leave `lxqt-config-brightness` as it is. Since we only support building with polkit turned on. And users who decide to compile it without the flag are prepared to have the functionality missing/not working.

Chiitoo:
Generally we do of course prefer to not carry patches for anything, and stay with what upstream supports as it makes packages a lot easier to maintain downstream, and upstreams won't get weird bug reports from set-ups they don't support/test.  Mainly it's up to the maintainer of the particular package though.

> I think in the Gentoo world people know that when they compile without certain
> use flags the program will lack certain functionality.

Most of the time... probably yeah.  :]  The flags can be quite overwhelming at first, but after fiddling with them for a bit, things become more obvious (I personally started using LInux with Gentoo, which was an adventure of its own (not that that adventure is over yet), and we do have perhaps surprisingly many such users visiting our forums).  Depending on the situation, there may be warnings displayed after installing a package, if some often wanted feature was disabled, to make things even more clear-like.

> I'm not sure about other DEs. But I guess that most (all?) require polkit and
> users who don't like it probably use a WM only.

Or the hack on the dependencies, heh.  :]

I've mainly used KDE, Enlightenment, Xfce, and now LXQt during my adventure, and I've managed without these packages so far.

If I recall correctly, during KDE 3 and 4 times, we had flags to make PolicyKit, ConsoleKit, and other friends like that optional, but with KDEPF 5 it's not as simple.  In fact, normally using just KWin for LXQt like I do, I'd already have polkit as a dependency due to Solid requiring udisks, which in turn pulls in polkit.  I get around this by simply removing the Solid dependency from udisks, and as such, avoid a bunch of things I don't need.

This might seem very weird and even silly to some, but I'm definitely not the only one out there, that likes to keep things like these off of their systems.  That is one of the reasons they enjoy Gentoo, as it basically just provides the tools (the package manager and the ebuild repository) for one to create their own distribution.

(I hope this doesn't seem too much like some sort of a promotion, but rather a bit of the reasoning behind making these things optional.  There are people who don't need them.)

Thank you everyone for your comments!

tsujan:
> but I'm definitely not the only one out there, that likes to keep things like these off of their systems. That is one of the reasons they enjoy Gentoo, as it basically just provides the tools

Before I read the word "Gentoo", I said, in my mind, "with most ditrsos we don't have easy ways of doing that -- maybe only with Gentoo". I hope Gentoo will continue this path; we need at least a no-nonsense distro ;)

Chiitoo:
Heh.  Yeah, I don't have much experience on other distributions at this time, but I am aware of that custom configurations are much more rare with them (for good reasons).  Of course one could build everything from source there, too, but probably without the help of a package manager.

As such, if/when I request something like this, I definitely consider the fact that what I'm asking might be a bit distro-specific, in that not many distributions are that likely to make use of it.

tsujan:
> but probably without the help of a package manager.

Arch is very flexible too but it prefers upstream.

PR #148, opened 2018-06-13
Backlight refactoring

mgerstner:
In the context of a code review of the `lxqt-backlight_backend` for openSUSE (see [bugzilla bug](https://bugzilla.suse.com/show_bug.cgi?id=1092192)) I want to suggest some improvements in the areas of security, cleanliness and small bugfixes.


jubalh:
Thank you for the changes @mgerstner.

GTM from me.

jubalh:
@luis-pereira, @palinek maybe you can review too.

lxqtbacklight/linux_backend/driver/libbacklight_backend.c
```diff
@@ -178,12 +175,13 @@ char *lxqt_backlight_backend_get_driver()
     char *driver;
     int n;
     char type[1024];
+    const char *sysfs_backlight_dir = "/sys/class/backlight";
```
palinek: How about `static`?

mgerstner: Yes good hint. I refactored to make this a global static constant. Also added another commit to avoid duplicate directory specification in arguments to `open_driver_file()`.

palinek:
Not related to this PR, but why are we using plain C here? @lxqt/lxqt

yan12125:
+1 for switching to C++. Even simply renaming the .c file to .cpp is a good move (stricter type checking, ...)

jubalh:
I wondered about the same. However let's discuss this in https://github.com/lxqt/lxqt/issues/1502

luis-pereira:
@palinek Don't know. It could have been done in C++ or even in Qt using `QFile`, `QTextStream` and `QString::asprintf()`. The performance penalty would be negligible and would be easier to avoid buffer overflows. Although I have nothing against C. Maybe @selairi and/or @tsujan can bring some light.

tsujan:
@luis-pereira I guess @selairi had written it in C and used it here. I have nothing against C at all.

jubalh:
Can we please move such discussion to https://github.com/lxqt/lxqt/issues/1502 ?
And discuss here the current PR?

lxqtbacklight/linux_backend/driver/libbacklight_backend.c
```diff
@@ -179,11 +191,10 @@ char *lxqt_backlight_backend_get_driver()
             if( !strcmp(dp->d_name, ".") || !strcmp(dp->d_name, "..") )
                 continue;
             driver = dp->d_name;
-            sprintf(path, "/sys/class/backlight/%s/type", driver);
-            FILE *in = fopen(path, "r");
+            FILE *in = open_driver_file("type", driver, "r");
             if( in == NULL )
                 continue;
-            int ok = fscanf(in, "%s", type);
+            int ok = fscanf(in, "%1024s", type);
```
luis-pereira: The documentation states that:
```
s      Matches  a  sequence  of  non-white-space  characters; the next pointer must be a pointer to
the initial element of character array that is long  enough to hold the input sequence and the
terminating null byte ('\0'), which is added automatically. 
```
so in order to accommodate the null byte shouldn't one write:
``` cpp
int ok = fscanf(in, "%1023s", type);
```
?

mgerstner: True. The exact section from `man 3 fscanf` that explains it is this one:

```
An optional decimal integer which specifies the maximum field width. [...]. String
input conversions store a terminating null byte ('\0') to mark the end of the
input; the maximum field width does not include this terminator.
```
I will change it to `%1023s`

mgerstner: I rebased the PR accordingly.

lxqtbacklight/linux_backend/driver/libbacklight_backend.c
```diff
@@ -67,15 +68,16 @@
 #define True 1
 #define False 0
 
-static int read_backlight(char *driver);
-static int read_max_backlight(char *driver);
-static int read_bl_power(char *driver);
+static FILE* open_driver_file(const char *file, const char *driver, const char *mode);
```
yan12125: A minor point: `FILE *open_driver_file` looks more "C" than `FILE* open_driver_file`.

lxqtbacklight/linux_backend/driver/libbacklight_backend.c
```diff
@@ -132,25 +129,40 @@ static int read_int(char *path)
     return value;
 }
 
-static int read_backlight(char *driver)
+static FILE* open_driver_file(const char *file, const char *driver, const char *mode)
```
yan12125: Same here

PR #150, opened 2018-06-14
lxqt-backlight: Simplify string construction

luis-pereira:
Taking into account that the strings are Latin1 encoded.


tsujan:
GTM

luis-pereira:
@agaida This is a no brainer.

PR #157, opened 2018-06-27
Don't use automatic string conversions

luis-pereira:
Disables automatic conversions from 8-bit strings (char *) to unicode
QStrings.
Disables automatic conversion from QString to 8-bit strings (char *).
Disables automatic conversions from QByteArray to const char * or const
void *.
Disables automatic conversions from QString (or char *) to QUrl.
Use QStringBuilder for more efficient string creation.

It make us aware of string and encoding conversions.


agaida:
Looks nice - what about adding these things to our build tools later?

luis-pereira:
@agaida Sure, After all repos are done.

PR #170, opened 2018-08-26
Position, alignment and size fixes for HTML delegate

tsujan:
Closes https://github.com/lxqt/lxqt/issues/1550

Texts and icons are vertically centered and item heights are correctly calculated in terms of wrapped text heights and icon heights. Unconditional using of `QTextDocument::adjustSize()` was wrong (actually, it could be removed). Moreover, `QStyle::drawControl()` should not be used for drawing the panel.

Please also note that Qt has an HTML bug, which is explained in a code comment.

### DIFF
--- a/lxqthtmldelegate.cpp
+++ b/lxqthtmldelegate.cpp
@@ -29,6 +29,7 @@
 #include "lxqthtmldelegate.h"
 #include <QAbstractTextDocumentLayout>
 #include <QTextDocument>
+#include <QApplication>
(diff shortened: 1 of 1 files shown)


tsujan:
Thanks for testing.

From your screenshot, I can see that Breeze has large layout spacings; they've made the "Panel" page too high. It isn't related to this PR but we might need to put the icon overriding setting in a separate section of this dialog.

PR #175, opened 2018-09-27
lxqtscreensaver: Postpone emitting of signals

palinek:
ref lxqt/lxqt#1515


palinek:
This doesn't work.... the whole X does not respond to application while locked....

PR #187, opened 2018-11-07
Fix config dialog cell height

tsujan:
Closes https://github.com/lxqt/lxqt/issues/1510


agaida:
GTM

tsujan:
And I was wrong because the next PR is in `lxqt-config` (last night was a very busy night). The comment can be read inversely (subtract X from Y instead of Y from X) and isn't important.

PR #198, opened 2019-01-07
Added missing text color of our HTML delegate

tsujan:
Fixes https://github.com/lxqt/lxqt/issues/1649

### DIFF
--- a/lxqthtmldelegate.cpp
+++ b/lxqthtmldelegate.cpp
@@ -98,13 +98,13 @@ void HtmlDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, 
     const QRect clip(0, 0, options.rect.width() - iconRect.width() - 8, options.rect.height());
     painter->setClipRect(clip);
 
-    // set text color to red for selected item
+    // set text colors
     QAbstractTextDocumentLayout::PaintContext ctx;
-    if (option.state & QStyle::State_Selected)
-    {
-        QPalette::ColorGroup colorGroup = (option.state & QStyle::State_Active) ? QPalette::Active : QPalette::Inactive;
+    QPalette::ColorGroup colorGroup = (option.state & QStyle::State_Active) ? QPalette::Active : QPalette::Inactive;
+    if (option.state & QStyle::State_Selected) // selected items
         ctx.palette.setColor(QPalette::Text, option.palette.color(colorGroup, QPalette::HighlightedText));
-    }
+    else // ordinary items and those with an alternate base color (there is no alternate text color)
+        ctx.palette.setColor(QPalette::Text, option.palette.color(colorGroup, QPalette::Text));
 
     ctx.clip = clip;
     doc.documentLayout()->draw(painter, ctx);


palinek:
The `::item` selector is still not considered with this patch.

tsujan:
I tested with this (and other variations) and it worked as intended:
```css
#commandList {
    background-color: #000000;
    alternate-background-color: #282828;
    selection-background-color: red;
    color: white;
    selection-color: black;
}

/*#commandList::item {
}*/

#commandList::item:selected {
    background: #e6e6e6;
    color: black;
}

#commandList::item:hover {
    background: rgba(255, 255, 255, 25%);
}

#commandList::item:selected:hover {
    background: white;
}

#commandList::item:alternate:hover {
    background: rgba(255, 255, 255, 12%);
}
```
Would you please be more specific?

tsujan:
And if you mean the text color of ordinary items, it's inside `#commandList {..}`, as that of `item:selected` is.

tsujan:
Explaining it in a better way:

Qt has no specific text color for view items. The ordinary items use `QPalette::Text` and the selected ones `QPalette::HighlightedText`. For that reason, the alternate base color should never have a high contrast with the base color.

palinek:
> Would you please be more specific?

Here with your modified .qss:
```css
#commandList {
    background-color: #000000;
    alternate-background-color: #282828;
    selection-background-color: red;
    color: white;
    selection-color: black;
}   

#commandList::item {
    color: red;
}

#commandList::item:selected {
    background: #e6e6e6;
    color: blue;
}

#commandList::item:hover {
    background: rgba(255, 255, 255, 25%);
}   

#commandList::item:selected:hover {
    background: white;
}   
    
#commandList::item:alternate:hover {
    background: rgba(255, 255, 255, 12%);
}
```

Text color does not follow the setting in `#commandList::item` and `#commandList::item:selected`

palinek:
> I explained why in my last comment. There is no `QPalette::ItemText`.

? This is what I'm getting with the above .qss:
- with `HtmlDelegate` in lxqt-runner
screenshot
- no `HtmlDelegate` in lxqt-runner
screenshot

tsujan:
In `HtmlDelegate`, we get colors only from QPalette.

palinek:
> In HtmlDelegate, we get text colors only from QPalette.

And that's the issue in reported in lxqt/lxqt#1649

tsujan:
Did text color work inside `#commandList {}` previously?

tsujan:
The main problem with styling Runner was that its item-view couldn't be styled correctly; I thought the reporter should be aware of that. Now, it can (the item and selected item text colors come from `#commandList {}`). I knew the text color couldn't be changed inside `::item:`when the colors are taken from `QPalette` but tested and saw it.

lxqthtmldelegate.cpp
```diff
@@ -98,13 +98,13 @@ void HtmlDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option,
     const QRect clip(0, 0, options.rect.width() - iconRect.width() - 8, options.rect.height());
     painter->setClipRect(clip);
 
-    // set text color to red for selected item
+    // set text colors
     QAbstractTextDocumentLayout::PaintContext ctx;
-    if (option.state & QStyle::State_Selected)
-    {
-        QPalette::ColorGroup colorGroup = (option.state & QStyle::State_Active) ? QPalette::Active : QPalette::Inactive;
+    QPalette::ColorGroup colorGroup = (option.state & QStyle::State_Active) ? QPalette::Active : QPalette::Inactive;
```
palinek: Shouldn't everywhere here be used `options` after `initStyleOption()`?

tsujan: We want the palette so, at least, it makes no difference.

BTW, the whole code was copy/pasted from an answer in stackoverflow -- half correct half not -- and apparently, no one had time to clean it up.

agaida:
@tsujan - is it ready?

tsujan:
> @tsujan - is it ready?

Yes, it is. It's the most that can be done and is enough for reasonable stylesheets.

If I find the time -- which isn't likely ;) -- I'll add matching Runner styles to LXQt themes.

P.S. You could reopen https://github.com/lxqt/lxqt/issues/1649 after this is merged but the *exact* issue mentioned there will never be fixed, IMO.

PR #202, opened 2019-01-30
Enable/disable a standard button if it exists

tsujan:
This is needed for adding Apply button later because Apply button should be enabled/disabled appropriately. Of course, it might find other usages too.


palinek:
Gtm

PR #212, opened 2019-03-31
Update lxqtbacklight_backend.c

damianatorrpm:
adds 3 cmd options (I use them to pipe the output as the original show method does not exactly what it says / I need)
```
    } if( !strcmp(argv[n], "--show-max") ) {
            show_max();
            return 0;
        } if( !strcmp(argv[n], "--show-value") ) {
            show_val();
            return 0;
        } if( !strcmp(argv[n], "--show-driver") ) {
```


luis-pereira:
@damianatorrpm lxqt-backlight_backend is a privileged executable (pkexec allowed without password).
I'm trying to keep it as simple as possible.
I'm not sure I understood what you are trying to achieve. Can you explain it in a more detailed manner ?

The format of the show command needs to be added. In this OR another PR.

PR #220, opened 2019-06-15
Make it build on macOS

yan12125:
* Don't use X11 libs.
* Don't build LXQt::ScreenSaver. It uses the X11 screensaver protocol.
* Use symbols from the KWindowSystem class instead of NET class. The
  NETWM header is available on XCB only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,11 +38,14 @@ set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTOUIC ON)
 
 find_package(lxqt-build-tools ${LXQTBT_MINIMUM_VERSION} REQUIRED)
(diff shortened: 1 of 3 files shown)


tsujan:
I don't have macOS and this isn't a review but just a question. Sorry if it doesn't make sense.

> Don't use KWindowSystem.

Wouldn't important functionalities be lost under macOS in this way? I don't remember the code but `KWindowSystem` seems too important to be ignored.

yan12125:
> Wouldn't important functionalities be lost under macOS in this way?

`LXQt::SingleApplication` will not work well. The window of the existing process will not be raised.

Thinking again - yes KWindowSystem should be enabled. Here is the updated pull request.

Note that vanilla KWindowSystem does not have a platform plugin for macOS. Patches like [1] are needed.

[1] https://github.com/RJVB/macstrop/tree/master/kf5/KF5-Frameworks/files/kwindowsystem

PR #226, opened 2019-08-03
Fix the wrong arguments in dbusCall

handsome-feng:


agaida:
@palinek @tsujan @luis-pereira  - i have no clue in depth so i can't judge

tsujan:
It's correct but doesn't have any consequence, `LXQT_SERVICE` and `LXQT_INTERFACE` being both "org.lxqt.session".

tsujan:
I meant GTM.

PR #227, opened 2019-08-18
lxqtscreensaver: allow custom lock command

yan12125:
The first step towards https://github.com/lxqt/lxqt/issues/1378.

### DIFF
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -28,6 +28,7 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include <QProcess>
+#include <QSettings>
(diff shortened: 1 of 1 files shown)


agaida:
@yan12125  - can it break something or can i simply merge?

yan12125:
Before this patch, users _might_ be able to guess what happens if the lock action fails as liblxqt shows different messages for different return values of `xdg-screensaver`. See https://github.com/lxqt/lxqt/issues/1378#issuecomment-516872970 for details.

After this patch, users need to run the command in the terminal to identify issues if something goes wrong. This is OK IMO.

agaida:
gtm

PR #242, opened 2019-12-13
Remove deprecated QDesktopWidget

luis-pereira:
It maintains functionality. It uses the default/primary screen.

### DIFF
--- a/lxqtpowermanager.cpp
+++ b/lxqtpowermanager.cpp
@@ -30,8 +30,8 @@
 #include <QDBusInterface>
 #include <QMessageBox>
 #include <QApplication>
-#include <QDesktopWidget>
 #include <QtDebug>
+#include <QScreen>
 #include "lxqttranslator.h"
 #include "lxqtglobals.h"
 #include "lxqtsettings.h"
@@ -80,7 +80,8 @@ protected:
     void resizeEvent(QResizeEvent* event) override
     {
         Q_UNUSED(event)
-        const QRect screen = QApplication::desktop()->screenGeometry();
+        const QScreen *primaryScreen = QGuiApplication::primaryScreen();
+        const QRect screen = primaryScreen ? primaryScreen->geometry() : QRect();
         move((screen.width()  - this->width()) / 2,
              (screen.height() - this->height()) / 2);
 


lxqtpowermanager.cpp
```diff
@@ -80,7 +80,7 @@ class LXQT_API MessageBox: public QMessageBox
     void resizeEvent(QResizeEvent* event) override
     {
         Q_UNUSED(event)
-        const QRect screen = QApplication::desktop()->screenGeometry();
+        const QRect screen = QGuiApplication::primaryScreen()->geometry();
```
tsujan: @luis-pereira 
To be exact, according to Qt → `qdesktopwidget.cpp`, it is so:
```c++
        auto screens = QApplication::screens();
        const QRect screen = screens.isEmpty() ? QRect() : screens.at(0)->geometry();
```

tsujan: Explanation:

In `qdesktopwidget.cpp`, the method `QDesktopWidgetPrivate::screen(int screenNo)` is used with 1 for `screenNo` in `QDesktopWidgetPrivate::screenGeometry(int screenNo)`:

```c++
QScreen *QDesktopWidgetPrivate::screen(int screenNo)
{
    QList<QScreen *> screens = QGuiApplication::screens();
    if (screenNo == -1)
        screenNo = 0;
    if (screenNo < 0 || screenNo >= screens.size())
        return nullptr;
    return screens.at(screenNo);
}
```

luis-pereira: @tsujan 
QGuiApplication::primaryScreen() already returns the  screen.at(0). But it also can return a null pointer. My code doesn't handles that.

A better solution is:
``` diff
-        const QRect screen = QApplication::desktop()->screenGeometry();
+        QScreen *primaryScreen = QGuiApplication::primaryScreen();
+        const QRect screen = primaryScreen ? primaryScreen->geometry() : QRect();
```

tsujan: > QGuiApplication::primaryScreen() already returns the screen.at(0)

You're right.

> A better solution is:

They're completely equivalent to each other.

luis-pereira: @tsujan I'm traveling and occupying the wait times with some boring tasks. I've written my comment without reading your second comment.

tsujan: >  I've written my comment without reading your second comment.

And your comment was correct :) Add a nullity check and merge it.

Safe journey!

luis-pereira: > > 
> 
> And your comment was correct :) Add a nullity check and merge it.

``` cpp
const QRect screen = primaryScreen ? primaryScreen->geometry() : QRect();
```
already does it.
> 
> Safe journey!

Thanks

luis-pereira:
Updated.

PR #246, opened 2020-01-30
PowerManager: correct dialog location

stream009:
Dialog show up in weird place on multi monitor environment.
Current code (wrongly) adjust dialog's geometry in various location. But
in fact these code aren't necessary because QDialog automatically move
dialog to center of the screen. It also do the right thing on multiple
monitor environment (center dialog on monitor that the mouse cursor
currently is in).


luis-pereira:
Can't test it in a multiple monitor now. Should work as expected, tough.

luis-pereira:
Rebased and merged.

PR #247, opened 2020-02-20
lxqtscreensaver: Fix a memory leak

luis-pereira:
QMessageBox::exec() is blocking so, it's safe to delete it.

### DIFF
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -150,10 +150,10 @@ public:
 
 void ScreenSaverPrivate::reportLockProcessError()
 {
-    QMessageBox *box = new QMessageBox;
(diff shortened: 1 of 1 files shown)


lxqtscreensaver.cpp
```diff
@@ -168,6 +168,7 @@ void ScreenSaverPrivate::reportLockProcessError()
     }
     box->setText(message.arg(lock_command));
     box->exec();
+    delete box;
```
palinek: How about using smart pointer or event stack allocated variable?

luis-pereira: Well the end result is the same.
Heap is not needed here, stack is fine.

PR #261, opened 2020-06-17
Don't use deprecated QProcess methods

luis-pereira:
Since we have to split the command ourselves, checking for a empty command,
makes sense now.

### DIFF
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -41,6 +41,7 @@
 #include <QProcess>
 #include <QCoreApplication> // for Q_DECLARE_TR_FUNCTIONS
 #include <QX11Info>
+#include <QDebug>
 
 #include <X11/extensions/scrnsaver.h>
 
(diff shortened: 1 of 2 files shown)


tsujan:
> ... checking for a empty command...

Yes, that may be the downside but, I think, in the end, the new way is better -- had encountered it in FeatherPad and removed the warning by checking emptiness twice.

GTM.

tsujan:
Oh, wait a minute! `QProcess::splitCommand` is only in Qt 5.15 (and so, I didn't use it in my code). Therefore, a version check is needed here.

IMVHO, we need to be caution about Qt 5.15 warnings. Sometimes, they aren't worth the fix so soon.

luis-pereira:
> Oh, wait a minute! `QProcess::splitCommand` is only in Qt 5.15 (and so, I didn't use it in my code). Therefore, a version check is needed here.

Hummm, that escaped me.
Will update. 

> IMVHO, we need to be cautious about Qt 5.15 warnings. Sometimes, they aren't worth the fix so soon.

I only fix things that don't require Qt>5.10.
The deprecation stuff is so much that important warnings are overlooked.

tsujan:
> I only fix things that don't require Qt>5.10.

I know; I've seen it.

luis-pereira:
Updated.

lxqtscreensaver.cpp
```diff
@@ -249,8 +250,19 @@ QList<QAction*> ScreenSaver::availableActions()
...
+#else
+        QStringList args = QProcess::splitCommand(d->lock_command);
+        if (args.isEmpty()) {
+            qWarning() << Q_FUNC_INFO << "Empty screen lock_command";
+            return;
+        }
+        const QString program = args.takeFirst();
+        d->m_lockProcess->start(program, args, QIODevice::ReadWrite);
+    }
+#endif
```
tsujan, line 265: `#endif` and the line above it (`    }`) need to be swapped -- a headache with version checks that has happened to me too several times ;)

luis-pereira: Been there too.
It's time to rest ;)

tsujan:
GTM

PR #264, opened 2020-07-19
lxqtbacklight: read brightness instead of actual_brightness for raw drivers

### LINKED ISSUE #263: lxqt-backlight: Discrepancy in reading and writing of brightness
lxqt-backlight gets the brightness value from `/sys/class/backlight/*/actual_brightness` instead of `/sys/class/backlight/*/brightness`.

This is fine actually, but `amdgpu_bl0` returns the raw brightness value (0-65536) for `actual_brightness` while `brightness` accepts only values between zero to `max_brightness` which is 255. When lxqt-backlight writes the processed `actual_brightness` to `brightness`, the brightness gets ramped up to 100% as the processed raw values are often in the range of 10000 to 20000, which get clamped to 255 when written to `brightness`.

Interestingly, this behaviour is not observed when one sets a brightness value directly - it only occurs when one increases or decreases the brightness through `lxqt-config-brightness -i`/`-d`. `lxqt-config-brightness -s 20` and `pkexec lxqt-backlight_backend 20 %` does what one would expect: setting the brightness value to 20% of the maximum. This is because both of them do `level / 100 * max_brightness` instead of `actual_brightness` and hence is not affected by the bug.

##### Expected Behavior
The brightness value should be obtained from `/sys/class/backlight/*/brightness`.

##### Current Behavior
The brightness value is obtained from `/sys/class/backlight/*/actual_brightness`.

##### Possible Solution
For `raw` drivers, read from `actual_brightness` instead of `brightness`. I'm not sure about the other driver types.

##### Steps to Reproduce (for bugs)
1. Have a laptop with an integrated AMD GPU. Raven Ridge (Vega 7) in my case.
2. Attempt to increase/decrease the brightness value, and blind your eyes.

##### System Information
* Distribution & Version: Arch Linux (rolling)
* Kernel: 5.7.8
* Qt Version: 5.15.0-4
* libqtxdg Version: 3.5.0
* lxqt-build-tools Version: N/A
* liblxqt version: 0.15.1-1

hch12907:
Closes #263

Whether this is a correct fix remains to be seen, as currently all my Linux computers I have on my hand are using `raw` type drivers, and I have nothing else to test the fix with.

### DIFF
--- a/lxqtbacklight/linux_backend/driver/libbacklight_backend.c
+++ b/lxqtbacklight/linux_backend/driver/libbacklight_backend.c
@@ -70,9 +70,13 @@
 
 static FILE* open_driver_file(const char *file, const char *driver, const char *mode);
 static int read_backlight(const char *driver);
+static int read_actual_backlight(const char *driver);
 static int read_max_backlight(const char *driver);
 static int read_bl_power(const char *driver);
+static int read_type(const char *driver);
 static const char *sysfs_backlight_dir = "/sys/class/backlight";
+
+typedef enum {FIRMWARE, PLATFORM, RAW, OTHER, N_BACKLIGHT} BacklightTypes;
 
 int lxqt_backlight_backend_get()
 {
@@ -80,7 +84,12 @@ int lxqt_backlight_backend_get()
     if( driver == NULL ) {
         return -1;
     }
-    int value = read_backlight(driver);
+    int value;
+    if (read_type(driver) == RAW) {
+        value = read_actual_backlight(driver);
+    } else {
+        value = read_backlight(driver);
+    }
     free(driver);
     return value;
 }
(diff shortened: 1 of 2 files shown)


lxqtbacklight/linux_backend/driver/libbacklight_backend.c
```diff
@@ -165,7 +179,27 @@ static int read_bl_power(const char *driver)
...
 }
 
-typedef enum {FIRMWARE, PLATFORM, RAW, OTHER, N_BACKLIGHT} BackligthTypes;
+static int read_type(const char *driver)
+{
+    FILE *in = open_driver_file("type", driver, "r");
+    if( in == NULL ) {
+        return -1;
+    }
+    char type[1024];
```
hch12907, line 188: 1024 is probably overkill here, but the same value is used in `lxqt_backlight_backend_get_driver()` as well.

PR #268, opened 2020-10-26
LXQtTheme: Handle self assignment properly

luis-pereira:
Copy assignment operators must be protected against self-assignment.
Good practice.


tsujan:
GTM

PR #275, opened 2021-01-03
Fix dbus errors with invalid QVariant data sent over it

adjamhub:
Sending invalid QVariant over dbus let systemd provider to fail retrieving informations. journalctl list warnings on that calls. This (easy) patch solves this.


palinek (approved):
GTM

PR #278, opened 2021-01-09
lxqtsettings: Make GlobalSettings public (export symbols)

palinek:
..to make it possible to use new Qt5(onwards) signal/slot syntax.

Note: needed for e.g. lxqt/lxqt-powermanagement#227


tsujan (approved):
GTM

PR #279, opened 2021-01-11
load color palette from theme

antis81:


tsujan (changes requested):
→ https://github.com/lxqt/lxqt-themes/pull/24#issuecomment-757945945

PR #283, opened 2021-02-11
lxqtsingleapplication: Use "handle quit signals" constructor

palinek:
.. to be able to shutdown the application cleanly.


luis-pereira (approved):
GTM.

PR #285, opened 2021-03-01
readbility

luis-pereira:


configdialog/lxqtpageselectwidget.cpp
```diff
@@ -109,7 +109,7 @@ PageSelectWidget::PageSelectWidget(QWidget *parent) :
     setSpacing(2);
     setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Expanding);
     setWordWrap(true);
-    setDragEnabled(NoDragDrop);
+    setDragEnabled(NoDragDrop != 0u);
```
tsujan, line 112: Here, you've discovered another mistake in the code (this time, not by me). `NoDragDrop` is actually `QAbstractItemView::NoDragDrop`, which is just `0`. IMHO, there's no reason to use it in that way. Instead, we could write:

```c++
setDragEnabled(false);
```

lxqthtmldelegate.cpp
```diff
@@ -98,8 +98,8 @@ void HtmlDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option,
 
     // set text colors
     QAbstractTextDocumentLayout::PaintContext ctx;
-    QPalette::ColorGroup colorGroup = (option.state & QStyle::State_Active) ? QPalette::Active : QPalette::Inactive;
-    if (option.state & QStyle::State_Selected) // selected items
+    QPalette::ColorGroup colorGroup = (option.state & QStyle::State_Active) != 0u ? QPalette::Active : QPalette::Inactive;
+    if ((option.state & QStyle::State_Selected) != 0u) // selected items
```
tsujan, line 102: I really doubt that substitutions like the above ones could make the code more readable (there are some in other places too). I'm afraid I wanted to say the same thing about too much usage of `nullptr` but `0u` might feel strange to code readers — especially if they're contributors who want to preserve the code style.

That being said, if you prefer them, there will be no objection from me.

luis-pereira: Will come back to you on this.

PR #288, opened 2021-03-17
Remove deprecated functions. Also, a fix on adjustedPoint method

adjamhub:


lxqtrotatedwidget.cpp
```diff
@@ -269,7 +269,11 @@ void RotatedWidget::wheelEvent(QWheelEvent *event)
         return;
     cascadeCall = true;
 
+#if (QT_VERSION >= QT_VERSION_CHECK(5,14,0))
+    QWheelEvent contentEvent(adjustedPoint(event->position()), event->globalPosition(), event->pixelDelta(), event->angleDelta(), event->buttons(), event->modifiers(), event->phase(), false);
+#else
     QWheelEvent contentEvent(adjustedPoint(event->pos()), event->globalPos(), event->delta(), event->buttons(), event->modifiers(), event->orientation());
+#endif
```
tsujan, line 276: `adjustedPoint()` is a public method. Changing it may break some code.

Instead, if the wheel event is rewritten as:

```c++
#if (QT_VERSION >= QT_VERSION_CHECK(5,14,0))
    QWheelEvent contentEvent(adjustedPoint(event->position().toPoint()), event->globalPosition(), event->pixelDelta(), event->angleDelta(), event->buttons(), event->modifiers(), event->phase(), false);
#else
.....
```
the compiler warning should be silenced.

adjamhub: It's not just because of the compiler warning. adjustedPoint method clearly calculates a QPointF result. And every ctor there (QMouseEvent, QWheelEvent) neets a QPointF.
IMHO the code is better this way. And I could find another call at adjustedPoint() in all my lxqt repos. 
Can you please reconsider? 
Waiting for the response :)

tsujan: It's simple: A function that takes `QPointF` can also take `QPoint`. There's nothing wrong with that.

tsujan: > And I could not find another call at adjustedPoint() in all my lxqt repos.

OK, if that's the case, it can remain `QPointF`, as you made it here. Let me check it first.

tsujan: Sorry, I leave this to other LXQt devs because it's very difficult for me to check *every* LXQt component that depends on `liblxqt` to see whether this change breaks them or not.

adjamhub: > It's simple: A function that takes `QPointF` can also take `QPoint`. There's nothing wrong with that.

Yes, that's easy and sure. But the question is: is a QPoint everytime enough if it calls for a QPointF?

adjamhub: > Sorry, I leave this to other LXQt devs because it's very difficult for me to check _every_ LXQt component that depends on `liblxqt` to see whether this change breaks it or not.

Just search for "org:lxqt adjustedPoint in:file" in the searchbar?

tsujan: > But the question is: is a QPoint everytime enough if it calls for a QPointF?

Yes, it is enough for us.

> Just search for "org:lxqt adjustedPoint in:file" in the searchbar?

I wouldn't trust GitHub for that but OK, I'm merging it. Please be more cautious about changing public methods! Thanks.

tsujan (approved):
GTM.

Reminder: Test all components for at least one week, before the next release!

PR #293, opened 2021-04-24
Set Qt 5.15 as Qt minimum version

adjamhub:
Removing moc include, as we "automoc", Also, calm down clang warning about lambda unused contests


tsujan:
Thanks but we already have an approved PR: https://github.com/lxqt/liblxqt/pull/292

Please always take a quick look at previous PRs before making one.

adjamhub:
oops, sorry :)

PR #310, opened 2022-05-04
Reversed case-insensitivity commit

tsujan:
This reverses https://github.com/lxqt/liblxqt/commit/ead8ceefa2a142b59e346b9df39a0070de709c73. After a discussion with @yan12125, it turned out to be redundant (especially see https://github.com/lxqt/lxqt-themes/issues/69#issuecomment-1114882618).


tsujan:
This is exactly gives 1.1.0. Merging...

PR #315, opened 2022-12-10
Add a timeout to the reboot/poweroff ... confirmation dialog

kreijack:
Add a timeout to the confirmation dialog for
- poweroff
- shutdown
- logout
- hibernate
- suspend If the timeout expires, it is considered like the user pressed Ok.

The timeout is set on the basis of the `timeout_confirmation` settings:
- -1   -> not timeout at all
- 0    -> the confirmation dialog disappear instantly
- 1... -> the dialog closes itself after the timeout set

This patch is inspired to the current behavior in mate-desktop, where the confirmation dialogs have a timeout.
Also this patch correct a bug where the method   `MessageBox::warning` ignores `title` and `text`

### DIFF
--- a/lxqtpowermanager.cpp
+++ b/lxqtpowermanager.cpp
@@ -33,19 +33,67 @@
 #include <QtDebug>
 #include <QScreen>
 #include <QFile>
+#include <QTimer>
+#include <QPushButton>
 #include "lxqttranslator.h"
 #include "lxqtglobals.h"
 #include "lxqtsettings.h"
 #include <XdgIcon>
 
+#define DEFAULT_TIMEOUT 10
+
 namespace LXQt {
 
-class LXQT_API MessageBox: public QMessageBox
+class LXQT_API MessageBoxWithTimeout: public QMessageBox
 {
-    Q_DECLARE_TR_FUNCTIONS(LXQt::MessageBox)
+    Q_DECLARE_TR_FUNCTIONS(LXQt::MessageBoxWithTimeout)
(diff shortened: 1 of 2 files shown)


stefonarch:
Gave a look (finally, sorry) at this, it's working fine. 


screen_area_sab_08:21:15_

This would need a PR to include in `lxqt-config-session` for
`timeout_confirmation=` if `leave_confirmation=true` 

@gfgit @tsujan should we include this in LXQt 2.0.0 ?

stefonarch:
Btw I think this message could be changed as it's not true anymore, with many applications which have crash memory. "Unsaved work could be lost..."

gfgit:
> Btw I think this message could be changed as it's not true anymore, with many applications which have crash recovery nowadays. "Unsaved work could be lost..."

Well it's worse to hope app recover and then they don't do it properly. At least user will double check.
To be honest other DE do not have this warning.

Anyway I like this change, only I don't have much time to review it now unfortunately.

tsujan:
> @tsujan should we include this in LXQt 2.0.0 ?

I don't like a timeout.

tsujan:
> I think this message could be changed as it's not true anymore, with many applications which have crash recovery nowadays.

It isn't only about crashes. An editor may or may not save changes on being terminated without prompt. Actually, some devs (like me) may believe that such a saving isn't good because there's no guarantee that the user wants to save a change to a document that he considers good, and saving to files with with strange names might be confusing.

PR #318, opened 2023-02-16
Typo fixes

isf63:
Typo fixes to comments/text and identifiers. Half of the changes are to CHANGELOG.

Affected files:
- CHANGELOG
- CMakeLists.txt
- lxqtapplication.h
- lxqtnotification.cpp
- lxqtnotification.h
- lxqtnotification_p.h
- lxqtsettings.cpp
- cmake/lxqt-config.cmake.in

### DIFF
--- a/lxqtapplication.cpp
+++ b/lxqtapplication.cpp
@@ -148,5 +148,5 @@ void Application::listenToUnixSignals(QList<int> const & signalList)
 
     if (SignalHandler::instance.isNull())
         SignalHandler::instance.reset(new SignalHandler{this, [this] (int signo) { Q_EMIT unixSignal(signo); }});
-    SignalHandler::instance->listenToSignals(signoList);
+    SignalHandler::instance->listenToSignals(signalList);
 }
--- a/lxqtapplication.cpp
+++ b/lxqtapplication.cpp
@@ -142,11 +142,11 @@ namespace
     QScopedPointer<SignalHandler> SignalHandler::instance;
 }
 
-void Application::listenToUnixSignals(QList<int> const & signalList)
+void Application::listenToUnixSignals(QList<int> const & signoList)
 {
     static QScopedPointer<QSocketNotifier> signal_notifier;
 
     if (SignalHandler::instance.isNull())
         SignalHandler::instance.reset(new SignalHandler{this, [this] (int signo) { Q_EMIT unixSignal(signo); }});
-    SignalHandler::instance->listenToSignals(signalList);
+    SignalHandler::instance->listenToSignals(signoList);
 }
--- a/lxqtapplication.cpp
+++ b/lxqtapplication.cpp
@@ -142,11 +142,11 @@ namespace
     QScopedPointer<SignalHandler> SignalHandler::instance;
 }
 
-void Application::listenToUnixSignals(QList<int> const & signoList)
+void Application::listenToUnixSignals(QList<int> const & signalList)
 {
     static QScopedPointer<QSocketNotifier> signal_notifier;
 
     if (SignalHandler::instance.isNull())
         SignalHandler::instance.reset(new SignalHandler{this, [this] (int signo) { Q_EMIT unixSignal(signo); }});
-    SignalHandler::instance->listenToSignals(signoList);
+    SignalHandler::instance->listenToSignals(signalList);
 }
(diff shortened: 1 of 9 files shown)


isf63:
@tsujan, could you review this old PR - asking because you merged most of my other typo fixes. Also I have two .ui/XML tab-order PRs, in lxqt-session and lxqt-config.

tsujan:
@isf63, thanks for the reminder.

Real life problems may interfere with reviewing (and coding, in general). Reminders like this are appreciated, when a PR was made before the previous release.

lxqtapplication.h
```diff
@@ -56,14 +56,14 @@ class LXQT_API Application : public QApplication
...
-     * \param handleQuitSignals flag if signals SIGINT, SIGTERM, SIGHUP should be handled internaly (\sa quit() application)
+     * \param handleQuitSignals flag if signals SIGINT, SIGTERM, SIGHUP should be handled internally (\sa quit() application)
      */
     Application(int &argc, char **argv, bool handleQuitSignals);
     ~Application() override {}
     /*! Install UNIX signal handler for signals defined in \param signalList
      * Upon receiving of any of this signals the \sa unixSignal signal is emitted
      */
-    void listenToUnixSignals(QList<int> const & signolList);
+    void listenToUnixSignals(QList<int> const & signalList);
```
tsujan, line 66: You forgot to change the argument name also in the definition of the function `Application::listenToUnixSignals`, near the end of `lxqtapplication.cpp`. (Please don't change anything else).

isf63: Yes, I was confused on that one, as the comment above the method declaration states `\param signalList`, yet elsewhere in `lxqtapplication.cpp` signo is used: `void listenToSignals(QList<int> const & signoList)`; inside `Application::Application(int &argc, char** argv, bool handleQuitSignals)`.

Perhaps change the comment and declaration param to `signoList`?

lxqtapplication.cpp
```diff
@@ -142,7 +142,7 @@ namespace
     QScopedPointer<SignalHandler> SignalHandler::instance;
 }
 
-void Application::listenToUnixSignals(QList<int> const & signoList)
+void Application::listenToUnixSignals(QList<int> const & signalList)
```
tsujan, line 145: Also, in the body of the function (this cannot be compiled).

isf63: Whoops :/

palinek: Btw. this wasn't any typo, but intentional name (signal numbers list) :-)

tsujan: > Btw. this wasn't any typo

The comment of the functions said `... defined in \param signalList...`.

"signolList" wasn't indicative of a number either. I think @isf63 was right to treat it as a typo, although a harmless one.

isf63: I'll change the identifier and comment to `signoList` as that is used elsewhere in lxqtapplication.cpp/h.

tsujan: > I'll change the identifier and comment to signoList as that is used elsewhere in lxqtapplication.cpp/h.

Oh, this is getting complicated for no good reason. That method is public and has a comment; "signoList" has no meaning as a name either. Your previous change was good, and I was going to merge it after a test....

isf63: `signoList` has a well-defined meaning - 'signo' in Unix/Linux means 'signal number'. Running `grep -o -i signo lxqtapplication.cpp` there were 13 instances of signo before the recent commit, including another signoList method param.

I agree, getting complicated, but I think it's done.

tsujan: > signoList has a well-defined meaning...

It may have been used by some devs and then copied. "Linux signo" leads to "Linux signal" in Google and Wikipedia. 

> there were 13 instances of signo

That's why I said, "That method is public and has a comment....". Anyway, I resigned from reviewing it. Hopefully, another dev will do it.

isf63:
Mostly typo corrections to comments.

Double checked and partially reviewed by @tsujan. Merging...

tsujan:
Please NEVER merge a PR that isn't approved by the reviewer(s).

tsujan:
Particularly in this case, which is about a fundamental library, merging without asking others could create bad headaches later. I reverted it in https://github.com/lxqt/liblxqt/commit/7b66ae5f8901fd69673fb3f03b0f83dfb96a8d4a

tsujan:
A suggestion: You might want to make another PR that includes only the typos that are NOT inside the codes.

PR #328, opened 2023-07-23
Show error under wayland

stefonarch:
Fixes https://github.com/lxqt/lxqt-powermanagement/issues/342
The error message is shown when trying to lock screen from menu, runner or as action in powermanagement > lid > lock screen

### DIFF
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -190,9 +190,8 @@ void ScreenSaverPrivate::_l_lockProcess_errorOccurred(QProcess::ProcessError)
 }
 
 bool ScreenSaverPrivate::isScreenSaverLocked()
-
 {
-if (QGuiApplication::platformName() == QStringLiteral("wayland"))
+    if (QGuiApplication::platformName() != QStringLiteral("xcb"))
     {
          return false;
     }
--- a/lxqtscreensaver.cpp
+++ b/lxqtscreensaver.cpp
@@ -37,10 +37,9 @@
 #include <XdgIcon>
 #include <QMessageBox>
 #include <QAction>
-#include <QApplication>
 #include <QPointer>
 #include <QProcess>
-#include <QCoreApplication> // for Q_DECLARE_TR_FUNCTIONS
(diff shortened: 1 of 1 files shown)


lxqtscreensaver.cpp
```diff
@@ -191,6 +191,8 @@ void ScreenSaverPrivate::_l_lockProcess_errorOccurred(QProcess::ProcessError)
 
 bool ScreenSaverPrivate::isScreenSaverLocked()
 {
+  if (QGuiApplication::platformName() == QStringLiteral("xcb"))
+  {
     XScreenSaverInfo *info = nullptr;
```
tsujan, line 196: Since the indentation is equal to 4 spaces in this source, you could either add enough spaces to respect the code style or just use this at the start instead:

```c++
    if (QGuiApplication::platformName() != QStringLiteral("xcb"))
    {
         return false;
    }
```

stefonarch: Less lines the better, thanks!

tsujan: Since we use X11 functions below this, comparing with "xcb" would make more sense, as suggested above. "wayland" can be used where there are Wayland-specific codes.

BTW, you added a redundant empty line and forgot to indent the `if`.

stefonarch: Done hopefully.  I thought it would be more explicit about wayland.

tsujan: It seems good now.

But I still think we shouldn't introduce `QGuiApplication` into `LXQt::ScreenSaver` but in the apps which use it. My reason is that `LXQt::ScreenSaver` may not be used in a GUI app, regardless of how *we* use it.

@palinek, is `QGuiApplication` safe here?

palinek: By this usage, it should be perfectly safe as:
```  QString QGuiApplication::platformName()
  {
      return QGuiApplicationPrivate::platform_name ?
             *QGuiApplicationPrivate::platform_name : QString();
  } 
```
and the `QGuiApplicationPrivate::platform_name` is a static pointer to `QString`.

tsujan (approved):

PR #337, opened 2024-02-08
Wip Qt 6 port

gfgit:
Depends on [lxqt-build-tools#77](https://github.com/lxqt/lxqt-build-tools/pull/77)


tsujan:
Glad to see that you've participated in the Qt6 port.

stefonarch:
Nice! Maybe you've seen already https://github.com/orgs/lxqt/projects/3/views/1

stefonarch:
For testing  purpose I've a KDE Neon developer edition in Vbox:

> find_package(PolkitQt6-1 REQUIRED) # TODO: does not seem to be available yet

screen_area_gio_14:58:15_

gfgit:
Nice! I didn't notice. Now I've setup a GNOME Boxes VM with Neon Unstable which is a bit faster than VirtualBox.
Also there is X11 specific code about XScreenSaver and lock screen in general which I don't know how to port to Wayland. For now they are just skipped and a warning is emitted.
Are there any compositor-agnostic Wayland protocols about lockscreen?

tsujan:
> which I don't know how to port to Wayland.

Please don't include porting to Wayland in this PR, which is for porting to Qt6.

stefonarch:
I forgot to mention that this Vbox with Neon already blocked twice here, just editing files.

> Are there any compositor-agnostic Wayland protocols about lockscreen?

For the future: https://wayland.app/protocols/ext-session-lock-v1

gfgit:
> > which I don't know how to port to Wayland.
> 
> Please don't include porting to Wayland in this PR, which is for porting to Qt6.

Yes of course, but at least a warning line so it's easier to find for later wayland port?

tsujan:
> at least a warning line so it's easier to find for later wayland port?

Right. My comment was just a general remark :) Otherwise, your work has always been clean.

stefonarch:
Managed to compile it on Neon, after 2 blocks:
* set(KF5_MINIMUM_VERSION "6.0.0") for some reason on Neon the package` libkf6windowsystem-dev` and realted are something like v5.245
* Cmake doesn't check for libxss-dev  so if it's not installed it stops at 82% with error about missing `scrnsaver.h`

gfgit:
> Managed to compile it on Neon, after 2 blocks:
> 
>     * set(KF5_MINIMUM_VERSION "6.0.0") for some reason on Neon the package` libkf6windowsystem-dev` and related are something like v5.245
> 
>     * Cmake doesn't check for libxss-dev  so if it's not installed it stops at 82% with error about missing `scrnsaver.h`

Can we do something about this or is it just a Neon bug?

stefonarch:
I've no idea if they will update to 6.* the package versions at release, just changed the version to 5.2 in `CMakeList.txt`.

stefonarch:
Compiling I get
```
    -- Looking for IceConnectionNumber in ICE - found
CMake Error at /usr/local/share/cmake/lxqt2/lxqt2-config.cmake:59 (include):
  include could not find requested file:

    /usr/local/share/cmake/lxqt2/lxqt-targets.cmake
Call Stack (most recent call first):
  CMakeLists.txt:25 (find_package)

```
Changing to "lxqt2" at the end of `lxqt-config.cmake.in` solves:
`    include("${CMAKE_CURRENT_LIST_DIR}/lxqt2-targets.cmake")`

But both apps I try (globalkeys and lxqt-about) don't find `#include QDialog` or `QObject` at some point.

gfgit:
> Changing to "lxqt2" at the end of `lxqt-config.cmake.in` solves:

Fixed with latest commit.

> But both apps I try (globalkeys and lxqt-about) don't find `#include QDialog` or `QObject` at some point.

I get this error too but only some files of `global-keys` CMake project seems affected. Also auto completion and syntax highlighting in Qt Creator work only on some files which means it cannot parse include directives, in fact compilation fails in same places.

gfgit:
More precisely I can build `lxqt-globalkeys` CMake target but not `all` target because it fails to build `lxqt-globalkeys-ui`

stefonarch:
Actually the same goes for here too: translations aren't build without changing prefix to `*.ts` files. I made a test with resetting all to "lxqt" everywhere and translations get installed and  my strange errors about "QDialog not found"  and else where gone too.

gfgit:
What's the error?

tsujan:
I took a quick look at the changes, and they seemed all good. Also, the code was compiled without problem under Manjaro Testing (+ KF6). Will read it again soon. Thanks!

lxqtsettings.cpp
```diff
@@ -500,6 +501,10 @@ QString LXQtTheme::qss(const QString& module) const
  ************************************************/
 QString LXQtThemeData::loadQss(const QString& qssFile) const
 {
+    // TODO: original QRegExp, check new syntax and QRegExp::RegExp2 meaning
+    // QRegExp(QL1S("url.[ \\t\\s]*"), Qt::CaseInsensitive, QRegExp::RegExp2);
+    static QRegularExpression urlRegexp(QLatin1String("url.[ \\t\\s]*"), QRegularExpression::CaseInsensitiveOption);
```
tsujan, line 506: `static const`?

gfgit: Done. Could you confirm syntax?

tsujan: It's OK, thanks.

tsujan (approved):

PR #347, opened 2024-05-25
Replaced `QLatin1String` with `QLatin1StringView`

tsujan:
See https://doc.qt.io/qt-6/qlatin1string.html.

Also, `QStringLiteral` is used where a function is not overloaded to take `QLatin1StringView`. See the note below https://doc.qt.io/qt-6/qlatin1stringview.html#details for the reason.


luis-pereira (approved):
GTM.
Consider https://github.com/lxqt/liblxqt/pull/349.

PR #351, opened 2024-08-23
Fixed setting of item geometries in grid layout

tsujan:
With a horizontal layout, the item width was increased by one pixel when there was empty space, but that one pixel wasn't taken into account anywhere else — and similarly with a vertical layout. I removed it, and the occasional position shifts disappeared from task buttons of lxqt-panel.

In short, the original calculations inside `GridLayout::setGeometry` didn't seem correct to me.

Fixes https://github.com/lxqt/lxqt-panel/issues/2065

### DIFF
--- a/lxqtgridlayout.cpp
+++ b/lxqtgridlayout.cpp
@@ -615,11 +615,9 @@ void GridLayout::setGeometry(const QRect &geometry)
 
     const int cols = d->cols();
     int itemWidth = 0;
-    int widthRemain = 0;
     if (stretch_h && 0 < cols)
     {
-        itemWidth = qMin((geometry.width() + sp) / cols - sp, d->mCellMaxSize.width());
-        widthRemain = (geometry.width() + sp) % cols;
+        itemWidth = qMin((geometry.width() - (cols - 1) * sp) / cols, d->mCellMaxSize.width());
     }
     else
     {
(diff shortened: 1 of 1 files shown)


tsujan:
I didn't know whom I should ask for reviewing/testing, partly because the code was very old, but also because I didn't know who would be accessible or could reproduce the panel issue. @palinek, @stefonarch, ...?

stefonarch:
I think I can reproduce, didn't ever really noticed.

Before:
https://github.com/user-attachments/assets/4ac5c35d-b67c-4710-a534-d1c406c557d7

After:
https://github.com/user-attachments/assets/40f11750-3e9c-4377-9e00-16d43fa8a61f

tsujan:
> I think I can reproduce, didn't ever really noticed.

Yes, your first video shows the shift @isf63 and I were talking about, and It doesn't happen in your second video.

tsujan:
> Tested and works.

Thanks for testing.

> Didn't test with vertical panels however.

Its logic is exactly the same.

> Looking at the diff (which I hardly understand), it looks like moderate logic changes.

I guess the original coder believed that a grid consisted of horizontal and vertical lines with a thickness of one pixel. That assumption complicated the logic and caused the problem.

tsujan:
However I look at it, the old code could have been written only based on a wrong assumption. Merging...

PR #352, opened 2024-09-29
Issues 2124 allowing adding item before the first item in LXQt::GridLayout

kurokawachan:
The current implementation of LXQt::GridLayout does not allow us to fix issue  https://github.com/lxqt/lxqt-panel/pull/2125 which is changing the way new item inserting into StatusNotifierWidget, this would allow us to fix it

### DIFF
--- a/lxqtgridlayout.cpp
+++ b/lxqtgridlayout.cpp
@@ -244,11 +244,13 @@ GridLayout::~GridLayout()
  ************************************************/
 void GridLayout::addItem(QLayoutItem *item)
 {
-    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::BeforeFirstOfList){
+    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::BeforeFirstOfList)
+    {
         d_ptr->mItems.prepend(item);
         return;
     }
-    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::AfterLastOfList){
+    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::AfterLastOfList)
+    {
         d_ptr->mItems.append(item);
         return;
     }
--- a/lxqtgridlayout.cpp
+++ b/lxqtgridlayout.cpp
@@ -45,6 +45,8 @@ public:
     int mColumnCount;
     GridLayout::Direction mDirection;
 
+    GridLayout::ItemAppendingLocation mItemAppendingLocation;
+
     bool mIsValid;
     QSize mCellSizeHint;
     QSize mCellMaxSize;
@@ -104,6 +106,7 @@ GridLayoutPrivate::GridLayoutPrivate()
     mColumnCount = 0;
     mRowCount = 0;
     mDirection = GridLayout::LeftToRight;
+    mItemAppendingLocation = GridLayout::AfterLastOfList;
     mIsValid = false;
     mVisibleCount = 0;
     mStretch = GridLayout::StretchHorizontal | GridLayout::StretchVertical;
@@ -241,7 +244,14 @@ GridLayout::~GridLayout()
  ************************************************/
 void GridLayout::addItem(QLayoutItem *item)
 {
-    d_ptr->mItems.append(item);
+    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::BeforeFirstOfList){
+        d_ptr->mItems.prepend(item);
+        return;
+    }
+    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::AfterLastOfList){
+        d_ptr->mItems.append(item);
+        return;
+    }
 }
 
 
@@ -363,7 +373,29 @@ void GridLayout::setDirection(GridLayout::Direction value)
         invalidate();
     }
 }
+ 
+/************************************************
 
+************************************************/
+GridLayout::ItemAppendingLocation  GridLayout::itemAppendingLocation() const
+{
+    Q_D(const GridLayout);
+    return d->mItemAppendingLocation;
+}
+
+
+/************************************************
+
+************************************************/
+void GridLayout::setItemAppendingLocation(GridLayout::ItemAppendingLocation value)
+{
+    Q_D(GridLayout);
+     if (d->mItemAppendingLocation != value)
+     {
+         d->mItemAppendingLocation = value;
+         invalidate();
+     }
+}
 /************************************************
 
  ************************************************/
--- a/lxqtgridlayout.h
+++ b/lxqtgridlayout.h
@@ -67,6 +67,17 @@ public:
     };
     Q_DECLARE_FLAGS(Stretch, StretchFlag)
 
+     /**
+       This enum type is used to describe the location of the newly added item.
+      **/
+     enum ItemAppendingLocation
+     {
+         BeforeFirstOfList,    ///< The new Item is inserted before the first of the list.
+         AfterLastOfList     ///< The new item is inserted after the last of the list.
+     };
+ 
+
+
 
 
     /**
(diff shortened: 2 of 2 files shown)


tsujan (changes requested):
Thanks for trying to make `liblxqt` better.

The title and comment of a PR should be self-explanatory, such that it could be found easily in the commits history later. The title and comment really need to be changed here.

Names like `ItemAppendingLocation` or `BeforeFirstOfList` are not only too long but also confusing. You could use names like these instead:
```c++
enum itemsOrder
{
    firstToLast;   ///< The newest item comes last
    lastToFirst    ///< The newest item comes first
}

```
Please note that, since `LXQt::GridLayout` can be used anywhere, `firstToLast` is the natural order, and so, it comes before `lastToFirst`. 

In this PR, you've mostly respected the style of the original code, *except for how curly brackets were used*.

kurokawachan:
Hello @tsujan Thanks for the review, the issue with `itemsOrder` is that, `ItemAppendingLocation` would tell people that it would only affect the newly added items, it would not change the order of existing items. 

For the commit message what about
`
allowing adding item before the first item in LXQt::GridLayout
`
What about squash the commits
For the curly brackets, I would fix it.

tsujan:
Thanks for the quick response.

> it would not change the order of existing items

It *should* change the order of existing items.

Let's first see whether we're talking about the same thing. Suppose that the items `A`, `B`, and `C` are added in this same order. Then, if  `itemsOrder` is `firstToLast` (the natural order), we should see `ABC`, while if it's `lastToFirst`, we'll see `CBA`. Do you have something else in mind?

> For the commit message what about...

If we're talking of the same thing, a title like this sounds good to me: "**Allow reversal of the order of items in `LXQt::GridLayout`**". The PR comment could be "*This could be used, for example, to reverse the order of items in the Status Notifier plugin of lxqt-panel.*"

kurokawachan:
Hello @tsujan 

It would only be in 
`
GridLayout::addItem(QLayoutItem *item)
`

So, it would not affect existing items.

tsujan:
> So, it would not affect existing items.

That's an important problem that should be fixed after making the above-mentioned changes. Without solving it, this PR can't be used self-consistently anywhere, including Status Notifier.

For example, Status Notifier has a "plus" button that shows up when some of its items become hidden. We don't want to see that button between other items but only before or after all of them. We might be able to work around this problem in the code of Status Notifier itself, but we use workarounds only for issues in libraries that *aren't* ours; liblxqt belongs to LXQt.

lxqtgridlayout.cpp
```diff
@@ -45,6 +45,8 @@ class LXQt::GridLayoutPrivate
     int mColumnCount;
     GridLayout::Direction mDirection;
 
+    GridLayout::ItemAppendingLocation mItemAppendingLocation;
+
```
tsujan, line 49:
```c++
    GridLayout::ItemsOrder mItemsOrder;
```
without an empty line above it.

lxqtgridlayout.cpp
```diff
@@ -104,6 +106,7 @@ GridLayoutPrivate::GridLayoutPrivate()
     mColumnCount = 0;
     mRowCount = 0;
     mDirection = GridLayout::LeftToRight;
+    mItemAppendingLocation = GridLayout::AfterLastOfList;
```
tsujan, line 109:
```c++
    mItemsOrder = GridLayout::firstToLast;
```

lxqtgridlayout.cpp
```diff
@@ -241,7 +244,16 @@ GridLayout::~GridLayout()
...
+    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::BeforeFirstOfList)
+    {
+        d_ptr->mItems.prepend(item);
+        return;
+    }
+    if(d_ptr->mItemAppendingLocation == GridLayout::ItemAppendingLocation::AfterLastOfList)
+    {
+        d_ptr->mItems.append(item);
+        return;
+    }
```
tsujan, line 256: Please change these to:

```c++
    if (d_ptr->mItemsOrder == GridLayout::ItemsOrder::firstToLast)
    {
        d_ptr->mItems.append(item);
    }
    else
    {
        d_ptr->mItems.prepend(item);
    }
```

lxqtgridlayout.cpp
```diff
@@ -363,7 +375,29 @@ void GridLayout::setDirection(GridLayout::Direction value)
...
+************************************************/
+void GridLayout::setItemAppendingLocation(GridLayout::ItemAppendingLocation value)
+{
+    Q_D(GridLayout);
+     if (d->mItemAppendingLocation != value)
+     {
+         d->mItemAppendingLocation = value;
+         invalidate();
+     }
+}
```
tsujan, line 400: Please remove the above methods and add these after `void GridLayout::setStretch(Stretch value)`:

```c++
/************************************************

 ************************************************/
GridLayout::ItemsOrder  GridLayout::itemsOrder() const
{
    Q_D(const GridLayout);
    return d->mItemsOrder;
}


/************************************************

 ************************************************/
void GridLayout::setItemsOrder(GridLayout::ItemsOrder value)
{
    Q_D(GridLayout);
    if (d->mItemsOrder != value)
    {
        d->mItemsOrder = value;
        std::reverse(d->mItems.begin(), d->mItems.end());
        invalidate();
    }
}
```

lxqtgridlayout.h
```diff
@@ -67,6 +67,17 @@ class LXQT_API GridLayout: public QLayout
...
+       This enum type is used to describe the location of the newly added item.
+      **/
+     enum ItemAppendingLocation
+     {
+         BeforeFirstOfList,    ///< The new Item is inserted before the first of the list.
+         AfterLastOfList     ///< The new item is inserted after the last of the list.
+     };
+ 
+
+
```
tsujan, line 80: Replace the above lines with

```c++
    /**
      This enum type is used to describe the order of items.
     **/
    enum ItemsOrder
    {
        firstToLast,   ///< The newest item comes last
        lastToFirst    ///< The newest item comes first
    };
```

lxqtgridlayout.h
```diff
@@ -140,6 +151,22 @@ class LXQT_API GridLayout: public QLayout
...
+ 
+     /**
+      Sets the way we insert new items for this grid.
+ 
+     \sa  GridLayout::ItemAppendingLocation
+     **/
+     void setItemAppendingLocation(ItemAppendingLocation value);
+ 
+
+
```
tsujan, line 169: Finally, replace the above lines with:

```c++

     /**
     Returns the order of items.

     \sa  GridLayout::ItemsOrder
     **/
     ItemsOrder itemsOrder() const;

     /**
      Sets the the order of items.

     \sa  GridLayout::ItemsOrder
     **/
     void setItemsOrder(ItemsOrder value);


```

tsujan:
For me it was easier to make a superseding PR instead of mentioning the above changes, especially because they include the entire PR. But, as your started this, it might be better if you make those changes and complete the work here.

After that, I'll make a PR for lxqt-panel and replace https://github.com/lxqt/lxqt-panel/pull/2125 with it, because all lines of https://github.com/lxqt/lxqt-panel/pull/2125 need changes.

kurokawachan:
Hello @tsujan Thanks for the review. Here are somethings we need to discuss here. 

The pr I made, here we only discuss liblxqt, it could be used self-consistently. As the default behavior would insert new items to the back of list. And this behavior only change when the user ask it to. 

If we are to revert the display order of the whole list, I might start a new pr doing that, instead of working on this one. Following your code modification only changing the name of variables. It would not in fact changing the behavior into "reverting the display order of the whole list". 

If you agree that this pr would keep the liblxqt self-consistently, that means we might as well keep working on this one. which only change the the location of newly inserted items.  I don't think it would impact the user for lxqt-panel, it is, however, contradicting the "plus" button being before or after all of them. 

So, the question, is it a hard requirement that "plus" button has to be at the begin or last of the list?
If so, I might start a new pr reverting the list.

tsujan:
Superseded by https://github.com/lxqt/liblxqt/pull/353

PR #353, opened 2024-09-30
Support reversing the order of items in LXQt's grid layout

tsujan:


palinek (approved):
Clean & straightforward

PR #356, opened 2024-12-18
Disable autostart entries only for LXQt

tsujan:
Related to https://github.com/lxqt/lxqt-session/issues/563

NOTE: This should be used only with an upcoming `lxqt-session` patch that uses it.


stefonarch:
GTM, didn't find an issue with existing desktop entries.

tsujan:
This was what made sense to me, but I'm all ears if a better logic is suggested.

PR #372, opened 2026-01-05
Support logout from `SystemdProvider`

basil:
### Background

`lxqtpower` provides a provider mechanism. Providers are initialized in https://github.com/lxqt/liblxqt/blob/46711d90d337552539656da57a0d013e6ec6fc50/lxqtpower/lxqtpower.cpp#L39-L45 according to these rules:

- If `useLxqtSessionProvider` is true, then `LXQtProvider` is initialized. This communicates with `lxqt-session` via DBus.
- `SystemdProvider` is unconditionally initialized with lower precedence.

When power actions are performed, the providers are tried in order: https://github.com/lxqt/liblxqt/blob/46711d90d337552539656da57a0d013e6ec6fc50/lxqtpower/lxqtpower.cpp#L57-L81 So first `LXQtProvider` is tried (if `useLxqtSessionProvider` is true), and then `SystemdProvider` is tried. Thus we have established that `lxqtpower` does not depend on `lxqt-session`.

### Problem

When `SystemdProvider` is in use (i.e., when `lxqt-session` is not running), `lxqt-leave` shows a grayed-out logout button, but the shutdown and reboot buttons are not grayed out.

### Evaluation

`SystemdProvider::canAction` and `SystemdProvider::doAction` include case statements for `case Power::PowerShutdown` and `case Power::PowerReboot`, but not for `case Power::PowerLogout`: https://github.com/lxqt/liblxqt/blob/46711d90d337552539656da57a0d013e6ec6fc50/lxqtpower/lxqtpowerproviders.cpp#L404-L477

### Solution

As discussed in https://github.com/lxqt/lxqt-wayland-session/issues/105#issuecomment-3710858488, add the missing case, calling systemd's `TerminateSession` via DBus. This is the equivalent of `loginctl terminate-session` ([man page](https://www.freedesktop.org/software/systemd/man/latest/loginctl.html)) which is documented as follows:

> Terminates a session. This kills all processes of the session and deallocates all resources attached to the session.

### Implementation details

`XDG_SESSION_ID` is set by systemd-logind for a user session. If it's missing, we are not running inside a logind-managed session and consider the functionality unavailable. systemd does not offer a `CanLogout` or `CanTerminateSession` method analogous to `CanReboot`, `CanPowerOff`, etc. So we therefore check if we have an `XDG_SESSION_ID` and that the DBus service, object, and interface are present on the system bus, then consider the functionality available.

For actually logging out the user, `TerminateSession` takes a string ID rather than an `interactive` boolean like the `PowerOff` and `Reboot` methods, so the `dbusCallSystemd` function is not suitable. Instead, we modify the `dbusCall` method to support arbitrary arguments (with a default value of no arguments). If binary compatibility is a concern, I can leave the existing function signature alone and create a new signature for invoking DBus methods with arguments.

### Testing done

Tested in Fedora 43 under UWSM (without `lxqt-session` running). Before this PR, the logout button was grayed out. After this PR, it is not grayed out, and clicking it logs out the session.

### DIFF
--- a/lxqtpower/lxqtpowerproviders.cpp
+++ b/lxqtpower/lxqtpowerproviders.cpp
@@ -30,6 +30,7 @@
 #include "lxqtpowerproviders.h"
 #include <QDBusInterface>
 #include <QProcess>
+#include <QProcessEnvironment>
 #include <QGuiApplication>
 #include <QDebug>
 #include "lxqtnotification.h"
@@ -75,7 +76,8 @@ static bool dbusCall(const QString &service,
               const QString &interface,
               const QDBusConnection &connection,
               const QString & method,
-              PowerProvider::DbusErrorCheck errorCheck = PowerProvider::CheckDBUS
+              PowerProvider::DbusErrorCheck errorCheck = PowerProvider::CheckDBUS,
+              const QVariantList &args = QVariantList()
               )
 {
     QDBusInterface dbus(service, path, interface, connection);
@@ -92,7 +94,7 @@ static bool dbusCall(const QString &service,
         return false;
     }
 
-    QDBusMessage msg = dbus.call(method);
+    QDBusMessage msg = !args.isEmpty() ? dbus.callWithArgumentList(QDBus::Block, method, args) : dbus.call(method);
 
     if (!msg.errorName().isEmpty())
     {
@@ -403,6 +405,19 @@ SystemdProvider::~SystemdProvider() = default;
 
 bool SystemdProvider::canAction(Power::Action action) const
 {
+    if (action == Power::PowerLogout)
+    {
+        const QByteArray sessionId = qgetenv("XDG_SESSION_ID");
+        if (sessionId.isEmpty())
+            return false;
+
+        QDBusInterface dbus(QL1SV(SYSTEMD_SERVICE),
+                            QL1SV(SYSTEMD_PATH),
+                            QL1SV(SYSTEMD_INTERFACE),
+                            QDBusConnection::systemBus());
+        return dbus.isValid();
+    }
+
     QString command;
 
     switch (action)
@@ -443,6 +458,21 @@ bool SystemdProvider::canAction(Power::Action action) const
 
 bool SystemdProvider::doAction(Power::Action action)
 {
+    if (action == Power::PowerLogout)
+    {
+        const QByteArray sessionId = qgetenv("XDG_SESSION_ID");
+        if (sessionId.isEmpty())
+            return false;
+
+        return dbusCall(QL1SV(SYSTEMD_SERVICE),
+                        QL1SV(SYSTEMD_PATH),
+                        QL1SV(SYSTEMD_INTERFACE),
+                        QDBusConnection::systemBus(),
+                        QL1SV("TerminateSession"),
+                        PowerProvider::CheckDBUS,
+                        QVariantList() << QString::fromLocal8Bit(sessionId));
+    }
+
     QString command;
 
     switch (action)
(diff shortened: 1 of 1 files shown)


tsujan:
Logging out of a session is the session's job. Shutting down a system is a system job.

I'm talking as an LXQt dev, not as an LXQt representative:

Until now, your suggestions were about cluttering LXQt's codes for this or that app to work without an lxqt-session. Well, that's the problem of this or that app. LXQt has a session and handles it gracefully.

basil:
Thanks @tsujan! As I mentioned in an email to you, I'm happy to discuss this with you further privately if desired.

tsujan:
> I'm happy to discuss this with you further privately if desired.

This isn't a private matter ;)

basil:
Thanks @tsujan, this pull request isn't the right place to have a broader discussion about systemd support at a high level. You are welcome to respond to my points in https://github.com/lxqt/lxqt-session/issues/623 if you wish to continue the conversation publicly on GitHub. I remain available to discuss the design in an audio or video chat, which may be recorded and made publicly available afterward (if desired).

lxqtpower/lxqtpowerproviders.cpp
```diff
@@ -403,6 +405,19 @@ SystemdProvider::~SystemdProvider() = default;
 
 bool SystemdProvider::canAction(Power::Action action) const
 {
+    if (action == Power::PowerLogout)
```
palinek: I would prefer to put this into the pre-existing `switch`.

lxqtpower/lxqtpowerproviders.cpp
```diff
@@ -443,6 +458,21 @@ bool SystemdProvider::canAction(Power::Action action) const
 
 bool SystemdProvider::doAction(Power::Action action)
 {
+    if (action == Power::PowerLogout)
```
palinek: Can this also be put into the `switch` and even the unified `dbusCall` used for all cases afterwards?

palinek (commented):
This PR seems perfectly valid (and thoroughly explained :thumbsup:) and I agree with the logic (I'm just putting some nitpicking code-flow comments).

Ad changing the static function signature -> it is a private not-exposed symbol, so IMO we don't need to be too cautious here

lxqtpower/lxqtpowerproviders.cpp
basil: Done in commit d91ff3e

lxqtpower/lxqtpowerproviders.cpp
basil: > Can this also be put into the switch

Done in commit d91ff3e

> [Can] even the unified dbusCall used for all cases afterwards?

Possibly? From [the docs](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html):

> CanPowerOff(), CanReboot(), CanHalt(), CanSuspend(), CanHibernate(), CanHybridSleep(), CanSuspendThenHibernate(), CanSleep(), CanRebootParameter(), CanRebootToFirmwareSetup(), CanRebootToBootLoaderMenu(), and CanRebootToBootLoaderEntry() test whether the system supports the respective operation and whether the calling user is allowed to execute it. Returns one of "na", "yes", "no", and "challenge". 

The docs say nothing about the return value of the `Reboot`, `PowerOff`, `Suspend`, and `Hibernate` commands, so I'm not sure if it's safe to replace the calls to `dbusCallSystemd` (which has special handling for the string values "yes" and "challenge") with calls to `dbusCall` (which does not have such special handling). The existing behavior of the reboot, shutdown, suspend, or hibernate functionality could likely be cleaned up, but this code cleanup deserves to be implemented and tested as part of a separate PR rather than adding risk to this PR.

stefonarch:
@palinek is this good now?

palinek:
Seems OK. But to be honor, I haven't tested/run anything of these changes.

stefonarch:
> Seems OK. But to be honor, I haven't tested/run anything of these changes.

I've no interest in uwsm but I'll run it for some days on my setup and if I notice no issue I'll merge it.

PR #378, opened 2026-04-28
Read brightness instead of actual_brightness with amdgpu

### LINKED ISSUE #1137: lxqt-config-brightness only decreasing brightness
Hello,

I installed lxqt with a fresh install of endeavorOS today. I noticed that both XF86MonBrightnessDown and XF86MonBrightnessUp were decreasing my brightness. I tried both lxqt-config-brightness -i and lxqt-config-brightness --increase in terminal, which both decreased my display's brightness. Issue was not fixed by updating or restarting the system. Issue has persisted since initial installation with no additional configuration.

Please help!

stefonarch:
Fixes https://github.com/lxqt/lxqt-config/issues/1137

See also https://github.com/systemd/systemd/issues/16302

@selairi stated  in this file "never read from brightness" but here it looks mandatory. On other 2 intel laptops `brightness` and `actual_brightness` are identical
@palinek @luis-pereira ?

### DIFF
--- a/lxqtbacklight/linux_backend/driver/libbacklight_backend.c
+++ b/lxqtbacklight/linux_backend/driver/libbacklight_backend.c
@@ -26,10 +26,10 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 /********************************************************************************
- *    This library uses Linux /sys/class/backlight files to read and change the 
+ *    This library uses Linux /sys/class/backlight files to read and change the
  *    backlight level.
- *    
- *    If screen backlight can be controlled, the Linux kernel will show inside 
+ *
+ *    If screen backlight can be controlled, the Linux kernel will show inside
  *    /sys/class/backlight directory one or more directories. Each directory has
  *    got the following files:
  *        /sys/class/backlight/driver/max_brightness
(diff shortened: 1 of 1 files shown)


stefonarch:
Had a look at brightnessctl, it looks like reading by default `brightness`: https://github.com/Hummer12007/brightnessctl/blob/e70bc55cf053caa285695ac77507e009b5508ee3/brightnessctl.c#L286

tsujan:
I'm not knowledgeable in this area, but this makes perfect sense to me.

luis-pereira:
It makes sense.
Can't test it. All my AMD machines are desktops.

tsujan:
IMHO, it may be better to give it to git users for more tests. It seems to solve the mystery that's mentioned at https://github.com/lxqt/lxqt-config/issues/1137#issuecomment-3025559737. Any objection?

luis-pereira:
> IMHO, it may be better to give it to git users for more tests. It seems to solve the mystery that's mentioned at [lxqt/lxqt-config#1137 (comment)](https://github.com/lxqt/lxqt-config/issues/1137#issuecomment-3025559737). Any objection?

None.

PR #381, opened 2026-05-15
Apply LXQt CMake settings to all code

luis-pereira:
LXQt CMake Settings should be applied to all code. That translates to
including the CMake files before adding directories.

lxqtbacklight/linux_backend/driver was added before LXQtCompilerSettings
were included. That meant that the `lxqt-backlight_backend` target didn't
use LXQtCompilerSettings settings, e.g., CMAE_C_STANDARD.

strdup() and popen() are not part of C99. They are POSIX functions.


stefonarch:
GTM

PR #382, opened 2026-05-20
cmake remove nested project

luis-pereira:
- **Remove nested CMake project**
- **Remove not needed include dirs**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,6 @@ set(LXQT_MAJOR_VERSION 2)
 set(LXQT_MINOR_VERSION 4)
 #
 # Patch Version, belong *only* to the component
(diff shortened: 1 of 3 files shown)


stefonarch:
GTM
CMakeList.txt:
>LXQt is 0.13 - liblxqt is at patch version 0

luis-pereira:
> LXQt is 0.13 - liblxqt is at patch version 0

Didn't catch the meaning.

stefonarch:
https://github.com/lxqt/liblxqt/blob/12477db97f36693c05d318f3787507660cc24999/CMakeLists.txt#L21

could be removed ;)

luis-pereira:
> could be removed ;)

Done

PR #389, opened 2026-09-18
linux_backend: use QProcess instead of blocking popen/fclose for the backlight helper

bornmw:
## Problem

When a backlight is available, `lxqt-panel` completely freezes ~60 s after start:
the panel clock stops and all widgets become unresponsive (start menu, windows,
tray). The panel process is alive (0% CPU) but its main thread is parked in
`wait4()`.

## Root cause

`LinuxBackend::setBacklight()` opens the privileged helper via

    popen("pkexec lxqt-backlight_backend --stdin", "w")

and arms a 60 s timer whose slot `closeBacklightStream()` calls `fclose(stream)`
on the **GUI thread**. `fclose` on a popen'd stream is `pclose()`, i.e. a
**blocking `waitpid()`**. The `sh`/`pkexec` chain leaves the write end of the
pipe held open, so the leaf `lxqt-backlight_backend` never sees EOF on its
stdin, never exits, and the blocking wait never returns. The Qt event loop of
the whole panel dies.

Repro on the reporter's system: the frozen panel's only child was
`lxqt-backlight_backend --stdin` running as root (via pkexec); a plain
EOF-delivered stream lets the helper exit in <0.1 s, while keeping any stray
write end open keeps it alive indefinitely (verified both).

## Fix

Manage the helper with a `QProcess` started without a shell:

- `setBacklight()` starts `pkexec lxqt-backlight_backend --stdin` directly;
  writes are buffered by Qt, so nothing blocks on start.
- `closeBacklightStream()` now does `closeWriteChannel()` + `terminate()`:
  neither blocks the main thread. `QProcess` reaps the child through the event
  loop; the `finished()` slot releases the object, so a later `setBacklight()`
  transparently restarts the helper.
- A 2 s safety-net timer escalates to `kill()` if the helper lingers.
- The destructor reaps deterministically with a bounded `waitForFinished(1000)`
  (only at panel shutdown).

Net effect: the GUI thread can no longer be held hostage by the (root)
backlight helper. No changes to the C helper or its public C API
(`lxqt_backlight_backend_get_write_stream` is simply no longer called by
`LinuxBackend`).

## Testing

- Build: `cmake -S . -B build && make -C build` on Qt 6 (repo floor is 6.6).
- With a backlight present, move the brightness slider: brightness updates,
  `lxqt-backlight_backend` appears/disappears via `pkexec` as before.
- After ~60 s idle the helper exits; the panel remains fully responsive (no
  `wait4` on the main thread, verified via `cat /proc/<lxqt-panel-pid>/task/1/wchan`).

Related: https://github.com/lxqt/lxqt/issues/2913


stefonarch:
Fails to compile here:

```
[ 97%] Building CXX object CMakeFiles/lxqt.dir/singleapplicationadaptor.cpp.o
/home/stef/git/lxqt/liblxqt/lxqtbacklight/linux_backend/linuxbackend.cpp: In member function ‘virtual void LXQt::LinuxBackend::setBacklight(int)’:
/home/stef/git/lxqt/liblxqt/lxqtbacklight/linux_backend/linuxbackend.cpp:108:59: error: ‘ForwardedErrorOutput’ is not a member of ‘QProcess’
  108 |         backlightProcess->setProcessChannelMode(QProcess::ForwardedErrorOutput);
      |                                                           ^~~~~~~~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/lxqt.dir/build.make:624: CMakeFiles/lxqt.dir/lxqtbacklight/linux_backend/linuxbac
```

bornmw:
> Fails to compile here:

Fixed, sorry somehow some modifications were made after the PR was prepared for submission, and they were not tested. Now it should be good

stefonarch:
This breaks `lxqt-config-brightness` cli which is used by shortcuts.
`lxqt-config-brightness  -s 100` and `lxqt-config-brightness  -d` ecc do nothing anymore.

bornmw (commented):
**Review of this branch.**

Verified:
- Compiles with Qt 6.11 (full `liblxqt.so` build, clean link).
- Runtime mechanism exercised against the real `pkexec /usr/bin/lxqt-backlight_backend --stdin` chain on a live machine: values written while the helper was still `Starting` as well as while `Running` were applied to the backlight; `closeWriteChannel()` let the helper notice EOF on stdin and exit on its own (exit 0) without any blocking `wait4`/`pclose` on the GUI thread; no root helper process lingered afterward.

Issues found in `b1378c5` (fixed in the immediately following commit):

1. **First value dropped.** After `start()`, `state()` is `Starting`, but the write guard required `Running`, so the brightness value that triggered the spawn was never written. Now writes proceed on `Starting || Running` — QProcess buffers input until the child runs, so nothing is lost.
2. **Wrong-object cleanup.** `finished()` unconditionally ran `backlightProcess->deleteLater()`. If a new helper was spawned before the old one was reaped, the old process's pending `finished()` would delete the *new* running `QProcess`. Fixed by capturing the spawned pointer and releasing only it.
3. **Unrelated changes.** The commit had reworded the license header block in both files (including "Lesser General Public License" → "General Public License" in two spots) plus whitespace churn. Reverted; the diff vs `master` now touches only `linuxbackend.cpp` / `linuxbackend.h`.

bornmw:
I'm converting this one to Draft, as this PR may be too excessive
Will create another one with minimal changes

stefonarch:
→https://github.com/lxqt/lxqt/wiki/Contributing-code#usage-of-llms

